PR #1, opened 2019-08-11
Add commonjs build output

LarsDenBakker:
This adds a simple build command which only transforms esm to commonjs for use in node. 

I know you can use the experimental modules flag, but we can't ask our users to do this as well so we have to stick to commonjs in node for now. We could do it in our package if you don't want this here, but I think others will run into this as well if they want to use it.

`main` is pointing to the commonjs build, and `module` is pointing to the esm build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,19 +2,27 @@
   "name": "es-module-lexer",
   "version": "0.1.2",
   "description": "Lexes ES modules returning their import/export metadata",
-  "main": "lexer.js",
+  "main": "dist/lexer.js",
+  "module": "lexer.js",
   "scripts": {
     "test": "mocha -r esm -u tdd test/unit.js",
+    "build": "babel lexer.js --out-dir dist --plugins @babel/plugin-transform-modules-commonjs",
     "bench": "node -r esm bench"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,19 +2,27 @@
   "name": "es-module-lexer",
   "version": "0.1.2",
   "description": "Lexes ES modules returning their import/export metadata",
-  "main": "lexer.js",
+  "main": "dist/lexer.js",
+  "module": "lexer.js",
   "scripts": {
     "test": "mocha -r esm -u tdd test/unit.js",
+    "build": "babel lexer.js --out-dir dist --plugins @babel/plugin-transform-modules-commonjs",
```
guybedford: We should compile with the `strict` option to avoid the `__esModule` interop applying.

LarsDenBakker:
Fully on board with jumping ship when es modules land in node v10 :)

package.json
LarsDenBakker: Done. Out of curiousity, why do you want this? It

guybedford: It was that without `strict`, I've pushed up the plugin to avoid this too.

Issue #4, opened 2019-08-20
Syntax error when using a $ inside a template string


LarsDenBakker:
This throws an error: ``` `$` ```

guybedford:
Resolved in #7.

PR #6, opened 2019-08-20
fix template literals with dollar signs

LarsDenBakker:
FYI we are now using this to resolve bare imports in https://github.com/open-wc/open-wc/tree/master/packages/es-dev-server. 

I started using it at work on a large codebase, and this is the only bug that came up. Pretty good I think :)

Fixes https://github.com/guybedford/es-module-lexer/issues/4

### DIFF
--- a/lexer.js
+++ b/lexer.js
@@ -156,7 +156,7 @@ function tryParseImportStatement () {
 function tryParseExportStatement () {
   if (readPrecedingKeyword(i + 5) !== 'export' || readToWsOrPunctuator(i + 6) !== '')
     return;
-        
+
   let name;
   charCode = str.charCodeAt(i += 6);
   commentWhitespace();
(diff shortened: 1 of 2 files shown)


lexer.js
```diff
@@ -268,13 +268,11 @@ function commentWhitespace () {
...
-        templateDepth = ++braceDepth;
-        return;
-      }
+    if (charCode === 36/*$*/ && str.charCodeAt(i + 1) === 123/*{*/) {
+      charCode = str.charCodeAt(i += 2);
+      templateStack.push(templateDepth);
+      templateDepth = ++braceDepth;
+      return;
     }
     else if (charCode === 96/*`*/) {
```
guybedford: Thanks for the PR!

There's a slightly easier fix - to just change this to an `if` instead of an `else if`!

I've added that and the test case from this PR to the Wasm branch, hoping to release that tomorrow.

guybedford:
This has now been merged with the Wasm conversion. Will credit you in the changelog for this fix.

Issue #3, opened 2019-08-18
Experimental Wasm Conversion


guybedford:
While the times for parsing are great when warmed up in the optimizer, this project heavily relies on cold performance because this is boot-time code analysis.

This is exactly the sort of reason to look into using Wasm.

I've started a branch to investigate this further, hope to have some results this week.

LarsDenBakker:
This is pretty interesting, what language do you plan on writing it in?

guybedford:
It has to be very lightweight (under 8KiB), so C is the obvious choice.

Nearly finished the conversion, hopefully should have a PR up tomorrow.

Because it's an exact port the current bugs should still apply to it, so don't let this stop any fixes.

guybedford:
PR at https://github.com/guybedford/es-module-lexer/pull/7.

guybedford:
Released in 0.3.0

PR #7, opened 2019-08-20
Wasm implementation

guybedford:
This converts the project into C with Wasm, resulting in a self-contained 10KiB .js file (including the Wasm binary), which is 4KiB minified.

Resolves: #8, #4, #3.

Tests passing. The API remains identical such that this is a patch release.

The benchmarks show a 6x improvement on cold start, which is the major justification here for Wasm conversion, given that this project needs to run on bootstrap so cannot rely on speeding up after the first X megabytes of source transform runs - it needs to be fast from the start.

Once v8 optimizations kick in performance is roughly the same, although further optimization is likely possible on top of this work.

Before conversion:

```
Cold Run, All Samples
bench/samples/*.js (2195 KiB)
> 179ms

Warm Runs (average of 25 runs)
bench/samples/d3.js (508 KiB)
> 6.84ms
bench/samples/d3.min.js (274 KiB)
> 3.16ms
bench/samples/magic-string.js (35 KiB)
> 0.6ms
bench/samples/magic-string.min.js (20 KiB)
> 0ms
bench/samples/rollup.js (929 KiB)
> 8.64ms
bench/samples/rollup.min.js (429 KiB)
> 5.12ms

Warm Runs, All Samples (average of 25 runs)
bench/samples/*.js (2195 KiB)
> 24.24ms
```

After conversion:

```
Cold Run, All Samples
bench/samples/*.js (2195 KiB)
> 33ms

Warm Runs (average of 25 runs)
test/samples/d3.js (508 KiB)
> 5.44ms
test/samples/d3.min.js (274 KiB)
> 3ms
test/samples/magic-string.js (35 KiB)
> 0.16ms
test/samples/magic-string.min.js (20 KiB)
> 0ms
test/samples/rollup.js (929 KiB)
> 9.2ms
test/samples/rollup.min.js (429 KiB)
> 5.04ms

Warm Runs, All Samples (average of 25 runs)
bench/samples/*.js (2195 KiB)
> 21.84ms
```


LarsDenBakker:
This is really interesting. What does this do in terms of browser and node support?

guybedford:
All environments that support ES modules already support Web Assembly, so this works in all the same environments.

jimmywarting:
currently the module is inlined as base64... and you also must call init to wait for the wasm to finish...
would it be better to use top level await + import.meta.resolve instead?

```js
const url = import.meta.resolve('./simple.wasm')
// alt.
const url = new URL('./simple.wasm', import.meta.url).toString()

const mod = await WebAssembly.instantiateStreaming(fetch(url), { ... })
export default instance.exports
```

Issue #9, opened 2019-08-21
Syntax error when exporting a function


LarsDenBakker:
With 0.3.x I'm getting a syntax error on this:
```js
export function a() {

}
```

guybedford:
Thanks, these cases help a lot. PR created.

Issue #2, opened 2019-08-18
Syntax error when exporting destructured variables


LarsDenBakker:
This gives a syntax error, but it runs fine in chrome:
```js
export const { a, b } = foo;
```

guybedford:
It would be worth checking against the latest master as I've made some changes today.

Otherwise a fix is very welcome, I expect it is because it is not exiting the `export ...` parse at the position before `{` resulting in an unmatched brace error.

LarsDenBakker:
Yeah still giving an error, I'll see if I can make a fix.

guybedford:
Thanks, I would take a look at it but I'm currently working on https://github.com/guybedford/es-module-lexer/issues/3!

guybedford:
Fixed in #11.

PR #11, opened 2019-08-21
Fix function and class exports

guybedford:
Fixes https://github.com/guybedford/es-module-lexer/issues/9 and https://github.com/guybedford/es-module-lexer/issues/2.


guybedford:
This also includes the destructuring fix for #2.

PR #12, opened 2019-08-21
Add text-encoder import for node

LarsDenBakker:
`TextEncoder` is a global variable in node 12, but not in node 10. This adds a small build step to add the import for the node build.

### DIFF
--- a/build.js
+++ b/build.js
@@ -15,4 +15,7 @@ const minified = MINIFY && terser.minify(jsSourceProcessed, {
   }
 });
 
+const nodeHeader = `import { TextEncoder } from 'util';`;
+
 fs.writeFileSync('./dist/lexer.js', minified ? minified.code : jsSourceProcessed);
+fs.writeFileSync('./dist/lexer.cjs.js', `${nodeHeader}${minified ? minified.code : jsSourceProcessed}`);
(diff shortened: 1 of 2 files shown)


build.js
```diff
@@ -15,4 +15,7 @@ const minified = MINIFY && terser.minify(jsSourceProcessed, {
   }
 });
 
+const nodeHeader = `import { TextEncoder } from 'util';`;
```
guybedford: Did you mean to use `require` here?

LarsDenBakker: This is before it's run through babel. If at some point you decide to ship es module in node, this would still work.

build.js
```diff
@@ -15,4 +15,7 @@ const minified = MINIFY && terser.minify(jsSourceProcessed, {
...
+
 fs.writeFileSync('./dist/lexer.js', minified ? minified.code : jsSourceProcessed);
+fs.writeFileSync('./dist/lexer.cjs.js', `${nodeHeader}${minified ? minified.code : jsSourceProcessed}`);
```
guybedford: Ah I see, note that the dist folder is published, so this file should be outputted to `lib` instead to avoid publishing it.

guybedford:
I've included an approach for this in https://github.com/guybedford/es-module-lexer/commit/f263c8a0860e6a1fa35bdcb28d97ea37450fc7e6. Will release shortly.

PR #16, opened 2019-08-21
add reproduction for node 10 error

LarsDenBakker:
This adds a setup where I'm getting an error on node 10. On node 12 it works fine.

I couldn't get the tests running on node 10, so I just created a simple common js file.

I stripped the `sinon.js` more and more, and when the array length became `14056` the error disappeared. 

Run `node node-10-error.js` to see the error.


node-10-error.js
```diff
@@ -0,0 +1,16 @@
+const fs = require('fs');
+const path = require('path');
+const { parse } = require('./dist/lexer.cjs.js');
```
guybedford: This should be `./dist/lexer.cjs`.

Note the extension is `.cjs` not `.js` because this package has a package.json `"type": "module"` field so a `.cjs` extension is needed in line with the latest --experimental-modules advice.

guybedford:
Perhaps try this again with the `.cjs` file - otherwise you would have still been using the old build I think.

LarsDenBakker:
Ah, good point.

I still get the error, but now with less characters: 28113 is when the error starts, 28112 is fine.

guybedford:
Ok I finally managed to track this down with a fix in https://github.com/guybedford/es-module-lexer/commit/46f9e907f5590d28a12637d5e1503a34a1fffac4. Releasing now.

Issue #17, opened 2019-08-21
methods named import incorrectly detected 


LarsDenBakker:
Here I am again :)

```js
class A {
  import() {

 }
}
```

Is detected as a dynamic import rather than a method. The same goes for:

```js
const a = {
  import() {

 }
}
```

guybedford:
These cases are working fine for me in a unit test... can you share an exact replication?

guybedford:
(Existing tests for similar cases can be found at https://github.com/guybedford/es-module-lexer/blob/master/test/_unit.js#L140)

LarsDenBakker:
I added a failing test: https://github.com/guybedford/es-module-lexer/pull/18

PR #18, opened 2019-08-22
add failing test for dyamic import method

LarsDenBakker:
Failing test for https://github.com/guybedford/es-module-lexer/issues/17


guybedford:
Ah, thanks, I was copying the cases into an existing source where it wasn't the first import, so the bug wasn't coming up. Fixed in https://github.com/guybedford/es-module-lexer/pull/19.

PR #27, opened 2019-08-28
add failing test for a very long export

LarsDenBakker:
When a module has a lot of exports, it's causing a memory access error:
```
     RuntimeError: memory access out of bounds
      at wasm-function[3]:43
      at wasm-function[13]:484
      at wasm-function[12]:211
      at parse (dist/lexer.js:2:344)
      at Context.test (test/_unit.js:319:5)
      at process.topLevelDomainCallback (domain.js:126:23)
```

Interestingly, this only occurs when the export statement doesn't have a lot of other code with newlines preceding it. 

This is happening for example with this file: https://unpkg.com/@fortawesome/free-solid-svg-icons/index.es.js. It parses just fine as is, but after it's been compiled by babel everything is put on a single line. Then it throws a memory access error. The same thing happens when you take the original file, and delete everything except for the export statement at the bottom.


LarsDenBakker:
This gives the same error:
```js
  test('x', () => {
    parse('export { a = };');
  });
```

PR #32, opened 2019-08-29
add failing test for empty export

LarsDenBakker:
After 0.3.10, an empty export is causing a parse error.


guybedford:
Thanks, I've landed the test with the fix.

Issue #36, opened 2019-09-06
Full import/export statement tracking


guybedford:
If we want to support hosting of import/export statements (needed for TLA), we should extend the data associated with imports and exports to include two new properties - `ss` and `se` for _statement start_ and _statement end_.

In the case of statements like `export function x () {}` we can actually return `-1` for `se` since a hoisting operation would just need to remove the `export` keyword and move that to the hoisted section.

FredKSchott:
Hey Guy, we're really excited about using this lib as a replacement for our custom Babel-based import scanner in Snowpack. Unfortunately, full import statement tracking is required for our use-case. More info here, if you're curious: https://github.com/pikapkg/snowpack/issues/146

I'm happy to help implement this if it's straightforward and you're interested (/cc @alex-saunders who may be interested as well).

guybedford:
@FredKSchott supporting full import binding tracking could certainly be possible, and shouldn't have any technical blockers. This project tries to keep the code footprint down as much as possible to avoid unnecessary costs for es-module-shims, so depending on how much this increases the final file size it could even be sensible to fork to get exactly the API you need, if it turns out it adds more than a few hundred bytes to the output size.

FredKSchott:
Okay, thanks. I'll see how far I can get into this then. Would love your guidance on it, it's been ages since i've done C development like this. 

Are you alright if we tackle imports first? Grabbing the whole "export" statement feels like a very different can of worms since the variable declaration can be any value (vs. import statements being much easier to statically analyze). Not sure if I have the resources to do both.

FredKSchott:
My first approach to parsing the import statement would be:

- "statement end" is always the same of the current "end" value +/1 0 for import.meta, +1 for regular imports (for the end quote), and +2 for dynamic imports (for the end paren and quote). We don't need any new data to track this, just calculate it inside of `addImport()` based on `dynamic`, and then add it to the return object under `se`.
- We need to pass  `startPos` from `tryParseImportStatement()` as a new argument into `addImport()`. We then add it to the return object under `ss`. No new logic needed otherwise.
- `export from ...` just means that we need to add a `startPos` variable to `tryParseExportStatement()` as well, so that we can pass that through to `addImport()`.

@guybedford Does this approach make sense? Overall this only adds one `startPos` pointer to the export parser, 1 argument to `addImport()`, and then one block of logic to calculate statement end (I got it naively in a 5 line if-else, but assuming we might be able to simplify that even further).

[...]

PR #41, opened 2020-01-24
Add support for parsing import statement start & end

FredKSchott:
Resolves one half of #36

- Adds `statement_start` (`ss`) and `statement_end` (`se`) to the data object for standard imports.
- For dynamic imports & import.meta, both values return -1.


README.md
```diff
@@ -58,6 +58,9 @@ import { init, parse } from 'es-module-lexer/dist/es-module-lexer.js';
   // Returns "asdf"
   source.substring(imports[0].s, imports[0].e);
 
+  // Returns "import { a } from 'asdf';"
+  source.substring(imports[0].ss, imports[0].se);
```
FredKSchott: can't comment outside the diff, but below here I saw that you're using `d`  for what is now statement_start. This works because we were already storing this value as the flag value for whether this import is dynamic or not. 

```
// Returns "import /*comment!*/ ("
  source.substring(imports[1].d, imports[1].s);	  source.substring(imports[1].d, imports[1].s);
```

Now that we have a true statement start/end value for standard imports, should we convert `dynamic` to `true` (or some other const/enum value) and then include statement start/end for dynamic imports as well?

src/lexer.c
```diff
@@ -280,21 +282,21 @@ void tryParseExportStatement () {
       ch = commentWhitespace();
       if (ch == 'f' && str_eq3(pos + 1, 'r', 'o', 'm')) {
         pos += 4;
-        readImportString(commentWhitespace());
+        readImportString((char16_t*)startPos, commentWhitespace());
```
FredKSchott: Casting here to remove this error I was seeing on compilation: 

```
src/lexer.c:285:26: warning: passing 'const char16_t *' (aka 'const unsigned short *') to parameter
      of type 'char16_t *' (aka 'unsigned short *') discards qualifiers
      [-Wincompatible-pointer-types-discards-qualifiers]
        readImportString(startPos, commentWhitespace());
```

If there's a better solution that you'd prefer lmk

FredKSchott:
I was able to see and fix all compilation errors, but clang was having trouble actually building the thing. I think the make file is using flags that my version of clang doesn't support. 

If you can check this out locally and compile + confirm that tests are passing, that would be great. Otherwise, when I next have time I can try to fiddle with updating clang (can you let me know what version you use / is needed?).

Makefile
```diff
@@ -5,7 +5,7 @@ lib/lexer.wasm: src/lexer.h src/lexer.c
 	@mkdir -p lib
 	clang src/lexer.c --sysroot=../wasi-sdk-6/opt/wasi-sdk/share/wasi-sysroot -o lib/lexer.wasm -nostartfiles \
 	-Wl,-z,stack-size=8192,--no-entry,--compress-relocations,--strip-all,\
-	--export=parse,--export=sa,--export=e,--export=ri,--export=re,--export=is,--export=ie,--export=id,--export=es,--export=ee \
+	--export=parse,--export=sa,--export=e,--export=ri,--export=re,--export=is,--export=ie,--export=iss,--export=ise,--export=id,--export=es,--export=ee \
```
guybedford: `ss`, `se` for statement start and statement end, saves two bytes :P

README.md
guybedford: That could well make sense... although worth noting this is a breaking change. Could be better to just double up for now then make the break later on.

src/lexer.c
guybedford: I think it would be better to make the argument in `readImportString` const.

src/lexer.h
```diff
@@ -115,6 +124,20 @@ uint32_t is () {
 uint32_t ie () {
   return import_read_head->end - source;
 }
+// getImportStatementStart
+uint32_t iss () {
+  if (import_read_head->dynamic == STANDARD_IMPORT)
+    return import_read_head->statement_start - source;
+  else
+    return -1;
```
guybedford: So I think we can just remove this path and the doubling up should work fine?

Makefile
guybedford: Although in theory I should do this for the others...

README.md
FredKSchott: works for me, I'll leave that potential cleanup for a later PR.

src/lexer.c
FredKSchott: done

Makefile
FredKSchott: done

src/lexer.h
FredKSchott: sgtm, done

FredKSchott:
@guybedford Okay, comments responded to, but I cannot for the life of me get this compiling. I've tried on both my local mac (`ld: unknown option: -z`) and a clean Ubuntu machine on AWS (`../wasi-sdk-6/opt/wasi-sdk/share/wasi-sysroot/include/wasi/core.h:15:2: error: <wasi/core.h> is only supported on WASI platforms.`). I'm sure that I did something stupid installing the right version of clang/etc., but unfortunately I don't have any more time available to spend on this for the moment.

guybedford:
@FredKSchott thanks for trying, yes I know it's a pain! All the more reason to migrate this project to Ziglang as I say, in due course :)

The build ran smoothly with just one or two bugs that needed correcting. I've gone ahead and landed on master, will push out a release shortly. Hope it helps you there! I'm sure there will be more bugs in this project, so further contributions back in future would be appreciated - this isn't a huge priority for me so can't guarantee ongoing support, but will do all I can to facilitate any PRs or interested maintainers.

guybedford:
Released in 0.3.14.

FredKSchott:
Ah thanks for getting this over the line @guybedford. I'll start to integrate it into my project and let you know if anything comes up in the future. Happy to contribute some PRs for anything that I find if time allows.

Thanks for the great project/package!

PR #43, opened 2020-02-11
fix path to es module version in the readme

daKmoR:


README.md
```diff
@@ -33,10 +33,11 @@ const { init, parse } = require('es-module-lexer');
 })();
 ```
 
-An ES module version is also available from `dist/es-module-lexer.js`:
+An ES module version is also available from `dist/lexer.js`:  
+Note: This version will be automatically used in rollup/es-dev-server/...  
 
 ```js
-import { init, parse } from 'es-module-lexer/dist/es-module-lexer.js';
+import { init, parse } from 'es-module-lexer';
```
guybedford: Strictly speaking, this isn't quite right since this references the CJS version.

If we update the package.json to read:

```js
{
  "main": "./dist/es-module-lexer.cjs",
  "exports": {
    "require": "./dist/es-module-lexer.cjs",
    "import": "./dist/es-module-lexer.js"
  }
}
```

Then Node.js would use the ES module version when modules are in use, and this readme would then be strictly correct.

Let me know if you'd like to add that to the PR, otherwise I will merge it with that in due course.

daKmoR: > Strictly speaking, this isn't quite right since this references the CJS version.

in which usage implementation? for rollup and es-dev-server the "module" field in the package.json is all that is needed to prefer this package.

I thought node did the same? if it needs the `exports` then the relevant parts in the package.json would end up to be something like this right?

```
{
  "main": "./dist/lexer.cjs",
  "module": "./dist/lexer.js",
  "exports": {
    "require": "./dist/lexer.cjs",
    "import": "./dist/lexer.js"
  }
}
```

guybedford: Yes, this is for Node.js.

guybedford: (and that is correct)

daKmoR: updated the info and added the needed section to package.json 👍 

shall I squash the commits or will you do that anyway on merge?

guybedford:
Thanks, released in 0.3.15.

FredKSchott:
Love the idea behind this PR, but for Snowpack we just realized that this change now introduces a warning when the package is run on Node v13. That's really confusing for users, and might end up blocking us from merging the PR that replaces our slow Babel parser with this package. 

@guybedford how do you feel about reverting this change until that warning message is gone?

See issue filed with Node to confirm that this was expected behavior on their part: https://github.com/nodejs/node/issues/31819#issuecomment-586750779

daKmoR:
oh boy I should have left it with just the readme change 🙈

guybedford:
I've gone ahead and reverted all these package.json changes in 0.3.17.

PR #45, opened 2020-02-15
Update package exports

alex-saunders:
This PR updates the `package.json` exports field to include valid paths (starting with `./`) as currently the exports are not following the spec (defined here https://nodejs.org/api/esm.html#esm_package_exports) 

We're trying to use `es-module-lexer` in snowpack but the current exports are breaking on Node 13 (you can see the full issue here https://github.com/pikapkg/snowpack/pull/208) so hopefully this should fix that issue for us and others!

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "main": "dist/lexer.cjs",
   "module": "dist/lexer.js",
   "exports": {
-    "require": "dist/lexer.cjs",
-    "import": "dist/lexer.js"
+    "require": "./dist/lexer.cjs",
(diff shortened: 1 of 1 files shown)


guybedford:
Thanks @alex-saunders for the fix here! I've pushed out a 0.3.16, great to hear it's working for you otherwise. And can you believe I wrote those docs 🤦‍♂ .... (https://github.com/nodejs/node/blame/master/doc/api/esm.md#L229)

alex-saunders:
Ha! I'll excuse you this time around 😆

PR #51, opened 2020-05-26
Fix syntax error on parsing empty import strings

TomerAberbach:
Hey! Thanks for your work on this package!

The parser currently throws an error when encountering static imports from the empty string (e.g. `import ''`, `import ""`, `import x from ''`, etc.). I'm guessing this is unintentional, but just in case it isn't, I'll explain my reasoning for supporting parsing of empty import strings. Although both `import ''` and `require('')` throw an error at runtime, importing or requiring the empty string shouldn't be a parser error because a [JavaScript `ModuleSpecifier` can be any `StringLiteral`](https://www.ecma-international.org/ecma-262/#prod-ModuleSpecifier).

This pull request corrects the problem and adds tests for parsing static imports from the empty string. The issue was that the `readImportString` method moved the current parser character position forward by one before invoking either the `singleQuoteString` or `doubleQuoteString` parsing methods, both of which already begin by moving the current parser character position forward by one. This wasn't a problem for non-empty strings because the `readImportString` method recorded the position of the first character in the string anyway, and the `singleQuoteString` and `doubleQuoteString` methods do not accumulate characters, but just find the ends of string literals so skipping the first character was not problematic. However, it was problematic for the empty string because the empty string has no first character so the closing quote of the empty string was skipped and a syntax error occurred.

Hope this helps!


guybedford:
Good stuff, thanks for the PR! I think you're the first who actually worked out how to build this thing, so congrats on figuring that out too :)

TomerAberbach:
Haha thanks! It took me a while...

Issue #53, opened 2020-09-20
the `se` of dynamic import is always 0


csr632:
Test code:
```
const { init, parse } = require("es-module-lexer");
(async () => {
  // either await init, or call parse asynchronously
  // this is necessary for the Web Assembly boot
  await init;

  const [imports, exports] = parse(`
  function loadDynamic() {
    import('react-dom').then(({ render }) => {
      dynamicImport.value = isFunction(render) ? 'success' : 'fail'
    })
  }
  `);
  console.log(imports);
})();
```
Output:
```
[ { s: 39, e: 50, ss: 32, se: 0, d: 32 } ]
```

guybedford:
@csr632 yes dynamic import is an expression not a statement so this value holds no meaning. Instead use the usual start and end.

To be consistent we could just match it up to the same as the `e` value I suppose.

csr632:
@guybedford 
My use case is wraping the import promise value. For example, I want to replace `import('react')` with `import('react').then(m => m.default)`. So `se` would be very helpful.

guybedford:
@csr632 the `e` value should be the closing `)` of the expression though I think.

csr632:
I am confused with `e` and `se` after seeing your comment :)

This is the `d.ts` for es-module-lexer:
```
// Type definitions for es-module-lexer 0.3
// Project: https://github.com/guybedford/es-module-lexer
// Definitions by: Tiger Oakes <https://github.com/NotWoods>
// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped

export interface ImportSpecifier {
    /**
     * Start of module specifier
     * @example
     * const source = `import { a } from 'asdf'`;
     * const [imports, exports] = parse(source);
     * source.substring(imports[0].s, imports[0].e);
     * // Returns "asdf"
     */
    s: number;
    /**
     * End of module specifier
     */
    e: number;

    /**
     * Start of import statement
     * @example
     * const source = `import { a } from 'asdf'`;
     * const [imports, exports] = parse(source);
     * source.substring(imports[0].s, imports[0].e);
     * // Returns "import { a } from 'asdf';"
     */
    ss: number;
    /**
     * End of import statement
     */
    se: number;

    /**
     * If this import statement is a dynamic import, this is the start value.
     * Otherwise this is `-1`.
     */
    d: number;
}
```
[unpkg](https://unpkg.com/browse/@types/es-module-lexer@0.3.0/index.d.ts)

[...]

PR #55, opened 2020-09-30
add more explanation for imports

csr632:
clarify for https://github.com/guybedford/es-module-lexer/issues/53


guybedford:
Thanks for the PR!

PR #64, opened 2021-02-28
fix: n is exist in all ImportSpecifier

xiaoxiangmoe:
In all ImportSpecifier, `n` field exists, it will be `string` or `undefined`. 
So `'n' in imports[0]` will always be true.
So we should use `n: string | undefined` rather than `n?: string`.


xiaoxiangmoe:
@guybedford When will it release?

PR #70, opened 2021-05-26
Remove window !== undefined check

aomarks:
Web Workers don't have `window` defined, so this check was causing workers to error from calling an undefined `Buffer.from` function.

After this change, es-module-lexer can be used in web workers.

I couldn't think of a reason that `window` needed to be checked in addition to `atob`, but I might have missed something.

Note that Node 16 now supports `atob`, so this change will cause Node 16+ to use `Uint8Array.from(atob())` instead of `Buffer.from()`, which seems fine. (In some future version that assumes Node 16+, you could drop the check entirely).

### DIFF
--- a/src/lexer.js
+++ b/src/lexer.js
@@ -57,7 +57,7 @@ function copyLE (src, outBuf16) {
 let wasm;
 
 export const init = WebAssembly.compile(
-  (binary => typeof atob === 'function' ? Uint8Array.from(atob(binary), x => x.charCodeAt(0)) : Buffer.from(binary, 'base64'))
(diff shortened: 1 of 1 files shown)


guybedford (changes requested):
I think one of the goals here was not to use the `atob` approach in Node.js.

I think it might be better to do a `typeof Buffer !== 'undefined' ? new Buffer() : Uin8Array.from(...)` reodering of the condition.

aomarks:
> I think one of the goals here was not to use the `atob` approach in Node.js.
> 
> I think it might be better to do a `typeof Buffer !== 'undefined' ? new Buffer() : Uin8Array.from(...)` reodering of the condition.

Done.

I figured the check was because `atob` never used to exist in Node (it was just added in Node 16). But maybe Buffer is faster?

guybedford:
I think in Node.js Buffer will definitely be faster, this looks good - thanks!

PR #71, opened 2021-06-04
test: replace assert.equal into assert.strictEqual

ycjcl868:
image


guybedford:
Looks good, thanks for the PR.

Issue #73, opened 2021-06-14
Expose information about import assertions


LarsDenBakker:
I am looking into polyfilling the new upcoming module types (JSON, CSS, HTML etc.). To do this it would be great if we can find out in a lightweight way whether an import is done using an import assertion. Is this something that could be added to ex module lexer? 

For an import like this:

```js
import foo from 'bar' assert { type: 'json' } 
```

And this would tell use that the import assertion is JSON. 

The proposal for import assertions is here: https://github.com/tc39/proposal-import-assertions

guybedford:
@LarsDenBakker I think we should build this directly into the core project, just like SystemJS supports - https://github.com/systemjs/systemjs/blob/main/src/extras/module-types.js.

Steps:
1. Support import assertions in ES module lexer (let me work on a PR now).
2. Integrate the module-types logic into the main repo here, assuming the constructed stylesheets polyfill like SystemJS does.

What is the status on HTML modules these days?

LarsDenBakker:
Great to hear! Import assertions were stalling progress for a while, but JSON module is stage 3 and CSS modules is making progress https://groups.google.com/a/chromium.org/g/blink-dev/c/vCrJCQxNnzo/m/OAYI9cPWAQAJ?pli=1. I'm not sure about the current status of HTML modules

LarsDenBakker:
Is it possible to include the start and end position of the import assertion? For my case I'd have to remove the actual import assertion from the code. es-module-shims will have to do this as well I think?

guybedford:
@LarsDenBakker thanks for the suggestions - I've added these features in https://github.com/guybedford/es-module-lexer/pull/74, review would be very welcome there.

PR #74, opened 2021-06-14
feat: add support for import assertions

guybedford:
Adds support for parsing [import assertions](https://github.com/tc39/proposal-import-assertions).

Three variables are used - `"a"`, `"as"` and `"ae"`.

For static imports these correspond to:

* `"a"`: The start of the `"assert"` keyword in the static import statement
* `"as"`: The `{` start of the assertion
* `"ae"`: The `}` end of the assertion

For dynamic imports, `"as"` and `"ae"` are not used, instead only `"a"`:

* `"a"`: The index immediately after the `,` - so `{` in the dynamic import expression `import('specifier', {...})`
* `"e"`: Is updated to reference the end of the specifier portion before the `,` in the import expression, instead of the position before the closing `)`. This is effectively a breaking change to the dynamic import handling.
* `"ss"`: Remains but is now documented as the start of the dynamic `import(...` expression itself.
* `"se"`: Is updated to reference the end of the import expression (position of the closing `)`).

Resolves https://github.com/guybedford/es-module-lexer/issues/73.


LarsDenBakker:
Thanks for this!

idev1990:
import-assertionshttps://github.com/guybedford/es-module-lexer/pull/74#issue-920812687

PR #77, opened 2021-07-05
Fix unable to import in node 12.18 ESM

### LINKED ISSUE #54: node exports?
hey,

I remember we added and then reverted exports to the package.json... 
I'm not sure however what the reason was?

Right now I am trying to import es-module-lexer in a node 14+ type="module" package and I need to do
```js
import { init, parse } from 'es-module-lexer/dist/lexer.js';
```

would be nice if I could just do
```js
import { init, parse } from 'es-module-lexer';
```

this could be archived with an export map that may look something like this
```js
"exports": {
  ".": {
     "require": "dist/lexer.cjs",
     "default": "dist/lexer.js",
  }
}
```

so the question is could we add it? and if not why was that again 😅

marvinhagemeister:
This PR adds some fields to `package.json` to be able to use this library with native ESM mode in node 12.18 and upwards. Without these node fails with the following error:

```bash
(node:1744) ExperimentalWarning: The ESM module loader is experimental.
file:///opt/build/repo/packages/wmr/src/lib/transform-imports.js:1
import { parse } from 'es-module-lexer';
         ^^^^^
SyntaxError: The requested module 'es-module-lexer' does not provide an export named 'parse'
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:92:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:107:20)
    at async Loader.import (internal/modules/esm/loader.js:179:24)
```

Ran into this while working on https://github.com/preactjs/wmr/pull/715 .

Explanation of the fix:

```js
"exports": {
  // Main/default entry point
  ".": {
    "import": "./dist/lexer.js",
    // CommonJS entry point
    "require": "./dist/lexer.cjs"
  },
  // If any module wants to import `es-module-lexerpackage.json` for some reason
  // (grabbing package.version or something)
  "./package.json": "./package.json",
  // Old-style deep imports, like `require("es-module-lexer/dist/lexer.js")`
  "./": "./"
},
```

Fixes #54 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,14 @@
   "main": "dist/lexer.cjs",
   "module": "dist/lexer.js",
   "types": "types/lexer.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/lexer.js",
+      "require": "./dist/lexer.cjs"
+    },
+    "./package.json": "./package.json",
+    "./": "./"
+  },
   "scripts": {
     "test": "NODE_OPTIONS=\"--experimental-modules\" mocha -b -u tdd test/*.cjs",
     "build": "node --experimental-modules build.js && babel dist/lexer.js | terser -o dist/lexer.cjs",
--- a/package.json
+++ b/package.json
@@ -6,11 +6,9 @@
   "module": "dist/lexer.js",
   "types": "types/lexer.d.ts",
   "exports": {
-    ".": {
-      "module" "./dist/lexer.js",
-      "import": "./dist/lexer.js",
-      "require": "./dist/lexer.cjs"
-    }
+    "module" "./dist/lexer.js",
+    "import": "./dist/lexer.js",
+    "require": "./dist/lexer.cjs"
   },
   "scripts": {
     "test": "NODE_OPTIONS=\"--experimental-modules\" mocha -b -u tdd test/*.cjs",
--- a/package.json
+++ b/package.json
@@ -7,11 +7,10 @@
   "types": "types/lexer.d.ts",
   "exports": {
     ".": {
+      "module" "./dist/lexer.js",
       "import": "./dist/lexer.js",
       "require": "./dist/lexer.cjs"
-    },
-    "./package.json": "./package.json",
-    "./": "./"
+    }
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,6 +5,14 @@
   "main": "dist/lexer.cjs",
   "module": "dist/lexer.js",
   "types": "types/lexer.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/lexer.js",
+      "require": "./dist/lexer.cjs"
+    },
+    "./package.json": "./package.json",
+    "./": "./"
```
guybedford, line 14:
```suggestion
    ".": {
      "module" "./dist/lexer.js",
      "import": "./dist/lexer.js",
      "require": "./dist/lexer.cjs"
    }
```

I've personally come around on the idea of using "module" for bundlers here.

Also I prefer not to expose unnecessary files. I understand there were issues with package.json loading but a custom node_modules lookup and package.json loader is 5 lines of a while loop so I don't think we should bend the entire JS ecosystem to saving these lines for metadata approaches.

guybedford (approved):
Thanks for the PR.

I also just noticed this one myself. The issue here is Node.js 12.18 doesn't yet supported the named exports for CJS feature, and as a result the package must be imported via `import lexer from 'es-module-lexer'`.

Fixing this via conditional exports seems good to me.

package.json
marvinhagemeister: Good point! Agree, that it's pretty trivial to do a custom lookup function. Just committed your suggestions.

package.json
```diff
@@ -5,6 +5,13 @@
   "main": "dist/lexer.cjs",
   "module": "dist/lexer.js",
   "types": "types/lexer.d.ts",
+  "exports": {
+    ".": {
+      "module" "./dist/lexer.js",
+      "import": "./dist/lexer.js",
+      "require": "./dist/lexer.cjs"
+    }
```
guybedford, line 13: One other thing - when there's only a main export, it's possible to flatten exports down to a single map as sugar. Perhaps unnecessary but I like it :)

```suggestion
    "module" "./dist/lexer.js",
    "import": "./dist/lexer.js",
    "require": "./dist/lexer.cjs"
```

guybedford (commented):
Ok, merging and releasing now.

PR #78, opened 2021-07-05
Add missing typings for import assertions

marvinhagemeister:
Past PRs #74 and #75 added support for import assertions, but the TypeScript types haven't been updated yet. This PR addresses that.


guybedford:
Thanks!

PR #84, opened 2021-08-16
Use valid ES6 syntax for catching the error

MrBananaLord:
It's a minor change and it's fixing the syntax, so it's a valid ES6. (It led to issue with asset precompilation when using more strict syntax checks.)


guybedford:
Thanks, happy to include that!

PR #86, opened 2021-09-18
feat: asm.js build

guybedford:
This adds an asm.js build of es-module-lexer, built using Emscripten 1.40.1 that still supported asm.js builds.

This fixes the cold start problem we have with the JS build, while still fully supporting CSP.

The output size minified and brotli'd is 4.7KB, as opposed to 5.2KB for the JS and 4.5KB for the Wasm builds.

With optimizations turned up, performance is comparable to Wasm!


GeoffreyBooth:
> With optimizations turned up, performance is comparable to Wasm!

Do you have any numbers on this?

guybedford:
The numbers are in the readme perf diff - https://github.com/guybedford/es-module-lexer#benchmarks.

PR #88, opened 2021-09-28
fixup: asm.js memory management

guybedford:
Resolves https://github.com/guybedford/es-module-shims/issues/202.

Fixes memory management for asm.js ensuring that for large sources > 1MB memory is dynamically allocated, while still ensuring that by default we only allocate 1MB heap for the workflow to reduce memory requirements.

In addition the build steps have been simplified slightly @richardo2016 this may help answer some of your questions re the manual steps.


richardo2016:
> Resolves [guybedford/es-module-shims#202](https://github.com/guybedford/es-module-shims/issues/202).
> 
> Fixes memory management for asm.js ensuring that for large sources > 1MB memory is dynamically allocated, while still ensuring that by default we only allocate 1MB heap for the workflow to reduce memory requirements.
> 
> In addition the build steps have been simplified slightly @richardo2016 this may help answer some of your questions re the manual steps.

That's cool! The changes of build steps show the manipulation to `src/lexer.js` before.

Issue #87, opened 2021-09-27
Cannot generate lib/lexer.js with emscripten 1.40.1


richardo2016:
I cannot build `lib/lexer.js` by myself with `make lib/lexer.js`,

Error like below:

```
$ make lib/lexer.js
emcc ./src/lexer.c -o lib/lexer.js -s WASM=0 -Oz --closure 1 \
        -s EXPORTED_FUNCTIONS=["_parse","_sa","_e","_ri","_re","_is","_ie","_ss","_ip","_se","_ai","_id","_es","_ee","_f","_setSource"] \
        -s ERROR_ON_UNDEFINED_SYMBOLS=0 -s SINGLE_FILE=1 -s TOTAL_STACK=15312 -s --separate-asm
emcc: warning: --separate-asm works best when compiling to HTML.  Otherwise, you must yourself load the '.asm.js' file that is emitted separately, and must do so before loading the main '.js' file. [-Wseparate-asm]
emcc: error: cannot --separate-asm when emitting wasm, since not emitting asm.js
make: *** [lib/lexer.js] Error 1
```

---

**Env Info**

OS: macOS
emscripten version: 1.40.1

as mentioned #86 , I installed emscripten 1.40.1 and activate it in my bash session.

```
$ ./emsdk install 1.40.1
$ ./emsdk activate 1.40.1
$ source ./emsdk_env.sh
```

I'm sure I haved installed it, because I checked it by `./emsdk list` and `./emcc -v`.

```
$ ./emsdk list
./emsdk list

The *recommended* precompiled SDK download is 2.0.30 (c69458f1bbf3ef5b8da4e934de210659cc9bca04).

To install/activate it, use one of:
         latest                  [default (llvm) backend]
         latest-fastcomp         [legacy (fastcomp) backend]

Those are equivalent to installing/activating the following:
         2.0.30             
         2.0.30-fastcomp    

All recent (non-legacy) installable versions are:
         2.0.30    
         2.0.29    
         2.0.28    
         2.0.27    
         2.0.26    
         2.0.25    
         2.0.24    
         2.0.23    
         2.0.22    
         2.0.21    
         2.0.20    
         2.0.19    
         2.0.18    
         2.0.17    
         2.0.16    
         2.0.15    
         2.0.14    
         2.0.13    
         2.0.12    
         2.0.11    
         2.0.10    
         2.0.9    
         2.0.8    
         2.0.7    
         2.0.6    
         2.0.5    
         2.0.4    
[...]

PR #89, opened 2021-09-28
feat: string-literal imports/exports for esmodule

richardo2016:
This PR is trying to support string-literal imports/exports for esmodule.

See proposals detail: https://github.com/bmeck/proposal-arbitrary-module-namespace-identifiers

Please @guybedford review changes of this PR to help make sure it compliance with convention.

---
[guybedford/cjs-module-lexer]:https://github.com/guybedford/cjs-module-lexer

**BTW**, I have tried to do some code synchronization from [guybedford/cjs-module-lexer]  as mentioned [here](https://github.com/guybedford/es-module-lexer/issues/87#issuecomment-927600588), but I found it's HARD to synchronize due to code context changes mentioned [here](https://github.com/guybedford/es-module-lexer/issues/87#issuecomment-927601092). So I just implement the feature as it should be in this repository.


richardo2016:
Is that necessary to keep same logic in `./lexer.js` with `./src/lexer.c`?

src/lexer.c
```diff
@@ -257,22 +257,45 @@ void tryParseImportStatement () {
...
+    
+    case '{': {
+      if (openTokenDepth != 0) {
+        pos--;
+        return;
+      }
+      bool end_curly_bracket = false;
+      while (pos < end) {
+        ch = *pos;
+        if (ch == '}') {
```
guybedford, line 286: We need to ensure we parse the quote strings fully here - otherwise this will fail on:

```
import { asdf as "b} from 'wrong'" } from 'right';
```

src/lexer.c
```diff
@@ -361,10 +384,17 @@ void tryParseExportStatement () {
       ch = commentWhitespace(true);
       while (true) {
         char16_t* startPos = pos;
-        readToWsOrPunctuator(ch);
+
+        ch = readToWsOrPunctuatorOrQuote(ch);
```
guybedford, line 388: I think this will only ever match the first quote for a string like `export { "p" as x } from 'z'` right? If so we may be able to replace this with just:

```suggestion
        if (!isQuote(ch)
          ch = readToWsOrPunctuator(ch);
```

The thing is, in the quote branch above we should still be scanning through the quotes (like the previous comment), otherwise we may fail on edge cases like:

```js
export { "p as 'z' from 'asdf'" as "z'" } from 'asdf';
```

src/lexer.c
```diff
@@ -403,14 +433,83 @@ void tryParseExportStatement () {
 
 char16_t readExportAs (char16_t* startPos, char16_t* endPos) {
```
guybedford, line 434: The `readExportAs` scan is specifically designed to always be at the possible start position of `as ...`. As per the previous comment this invariant is being changed which opens up a class of bugs. We should make sure to keep this invariant.

src/lexer.c
```diff
@@ -403,14 +433,83 @@ void tryParseExportStatement () {
...
+
+    if (end_quote == '}') {
+      syntaxError();
+      return ch;
+    }
+
+    readToWsOrPunctuator(ch);
+    endPos = pos;
+    withAs = str_eq2(pos + 1, 'a', 's');
+  }
```
guybedford, line 465: This quote scanning should move up into its own function / into the parent, then we can just keep the as logic below.

src/lexer.c
```diff
@@ -403,14 +433,83 @@ void tryParseExportStatement () {
...
+
+      if (end_quote == '}') {
+        syntaxError();
+        return ch;
+      }
+      readToWsOrPunctuator(ch);
+      endPos = pos - 1;
+    } else {
+      endPos = pos;
+    }
```
guybedford, line 498: Again, sharing the quote logic in its own function might help simplify things.

guybedford (commented):
Thanks for the work on this. Unfortunately it still needs a little more work. Let me know if I can clarify further.

src/lexer.c
richardo2016: OK, I will consider recover it and use new process.

src/lexer.c
```diff
@@ -250,29 +250,65 @@ void tryParseImportStatement () {
...
+
+    case '{': {
+      // import statement only permitted at base-level
+      if (openTokenDepth != 0) {
+        pos--;
+        return;
+      }
+      char16_t firstQuote;
+      bool curlyBraceClosed = false;
+      bool quoteClosed = true;
```
guybedford, line 286: Neither of these carried states should be necessary - parse state should be implicit by where in the parsing function things are - not in a variable at all.

guybedford: If the `stringLiteral` function has finished - then the quote is closed. There's no variable there - it's just running the function.

src/lexer.c
```diff
@@ -250,29 +250,65 @@ void tryParseImportStatement () {
...
+          }
+
+          if (!firstQuote) {
+            firstQuote = ch;
+            quoteClosed = false;
+          } else if (ch == firstQuote) {
+            quoteClosed = true;
+          }
+        } else if (ch == '}') {
+          curlyBraceClosed = quoteClosed;
```
guybedford, line 303: Rather than handling this as a variable and keeping the loop going, just break the loop and have the `readImportString` check after this loop.

src/lexer.c
```diff
@@ -403,26 +475,61 @@ void tryParseExportStatement () {
...
+    else {
+      startPos = pos + 1;
+
+      if (!scanExportAsQuotes(ch)) {
+        syntaxError();
+        return ch;
+      }
+      readToWsOrPunctuator(ch);
+      endPos = pos - 1;
+    }
```
guybedford, line 504: Start pos and end pos here are only being used to check if there was an import string - and not for their actual positions, so there is no need to offset the quotes.

Suggestion:

```suggestion
    // export { mod as "..." } from
    if (isQuote(ch)) {
      stringLiteral(ch);
      ch = readToWsOrPunctuator(ch);
    }
    else {
      ch = readToWsOrPunctuator(ch);
    }
    endPos = pos;
```

src/lexer.c
```diff
@@ -403,26 +475,61 @@ void tryParseExportStatement () {
...
+  char16_t ch;
+  while (++pos < end) {
+    ch = *pos;
+    if (ch == quote && *(pos - 1) != '\\') {
+      return true;
+    }
+  }
+
+  return false;
+}
```
guybedford, line 524: We shouldn't need this function - we can just use `stringLiteral(quote)`.

src/lexer.c
```diff
@@ -361,10 +397,46 @@ void tryParseExportStatement () {
...
+          ch = commentWhitespace(true);
+
+          if (str_eq2(pos, 'a', 's')) {
+            startPos = pos;
+            endPos = pos;
+          } else {
+            startPos = posQuoteStart + 1;
+            endPos = posQuoteEnd;
+          }
+        }
```
guybedford, line 436: All of this should be able to copy the same sort of approach as:

```
// export { "mod" ...
    if (isQuote(ch)) {
      stringLiteral(ch);
      ch = readToWsOrPunctuator(ch);
    }
    else {
      ch = readToWsOrPunctuator(ch);
    }
    endPos = pos;
```

guybedford (commented):
Getting there. Don't worry too much about the JS build for now let's just get the parser working first.

src/lexer.c
richardo2016: So that's it (⊙o⊙)… On coding this part, I always want locate the "REAL" position, and I found it works as I **didn't** locate the real position in some cases 😂

src/lexer.c
richardo2016: Same as previous conversation, if we need keep the quotes in `export { foo0 as "~123" } from './mod0.js';`, we can just use your suggested logic, should we keep them?

src/lexer.h
```diff
@@ -199,19 +199,22 @@ bool parse (uint32_t point);
 void tryParseImportStatement ();
 void tryParseExportStatement ();
 
+bool scanExportAsQuotes(char16_t quote);
+
```
guybedford, line 203:
```suggestion
```

src/lexer.c
```diff
@@ -403,14 +473,34 @@ void tryParseExportStatement () {
 
 char16_t readExportAs (char16_t* startPos, char16_t* endPos) {
   char16_t ch = *pos;
+
   if (ch == 'a') {
     pos += 2;
     ch = commentWhitespace(true);
     startPos = pos;
```
guybedford, line 480: This line can move inside of the `if (!isQuote(ch))` condition.

src/lexer.c
```diff
@@ -403,14 +473,34 @@ void tryParseExportStatement () {
...
+    // export { mod as "space " } from
+    // export { mod as "not~id" } from
+    // export { mod as "%notid" } from
+    else {
+      startPos = pos + 1;
+      
+      stringLiteral(ch);
+      readToWsOrPunctuator(ch);
+      endPos = pos - 1;
+    }
```
guybedford, line 499: Perfect, yes it's this simple.

src/lexer.c
```diff
@@ -361,10 +397,44 @@ void tryParseExportStatement () {
...
+          ch = commentWhitespace(true);
+
+          if (str_eq2(pos, 'a', 's')) {
+            startPos = pos;
+            endPos = pos;
+          } else {
+            startPos = posQuoteStart + 1;
+            endPos = posQuoteEnd;
+          }
+        }
```
guybedford, line 434: This entire section should be able to simplify with just the change as below - all three specifier parse cases are just being replaced with this same branching, that's all this PR needs to do:

```
if (!isQuote(ch)) {
      ch = readToWsOrPunctuator(ch);
      endPos = pos;
    }
    // export { mod as "identifer" } from
    // export { mod as "@notid" } from
    // export { mod as "spa ce" } from
    // export { mod as " space" } from
    // export { mod as "space " } from
    // export { mod as "not~id" } from
    // export { mod as "%notid" } from
    else {
      startPos = pos + 1;

      stringLiteral(ch);
      readToWsOrPunctuator(ch);
      endPos = pos - 1;
    }
```

guybedford (commented):
Nearly there.

src/lexer.c
richardo2016: Let me try it, I'm afraid maybe there are some special edge cases 🤔

richardo2016: I haved tried, if keep `readExportAs` unchanged as main branch. We couldn't do like code above(I removed comment lines):

```
if (!isQuote(ch)) { // branch-1
      ch = readToWsOrPunctuator(ch);
      endPos = pos;
      commentWhitespace(true); // in fact, this line is required on assumption that this snippet is valid
    } else { // branch-2
      startPos = pos + 1;

      stringLiteral(ch);
      readToWsOrPunctuator(ch);
      endPos = pos - 1;
    }
```

for code `export { "@unicode" as identifer } from 'xxx'`, we would extract both `'@unicode'` and `'identifer'`:

1. first, parser meet `"@unicode"`, enter branch-2, startPos represents pos of `@`, after `stringLiteral(ch)` & `readToWsOrPunctuator(ch)`, endPos represents the **space before `as`**, in loop, `ch = readExportAs(startPos, endPos);` would **add export `@unicode`**, but that's **Incorrect** ⚠️
2. then, re-enter of loop, parser will go to **space after `as`**; re-enter loop, parse meet `identifier`, and extract it, this is correct, but we have mistake in step 1.

I have met this case on refactor this entries section, at your reminding, I considered ensuring `readExportAs` always deal with exact whole module name token like:

1. `"@unicode"` in `export { "@unicode" } from 'xxx'`
2. `as identifer` in `export { "@unicode" as identifer } from 'xxx'`
3. `as "@identifer"` in `export { "@unicode" as "@identifer" } from 'xxx'`

So maybe we should do a little bit more changes here on top of main branch, I have pushed my latest change @guybedford 

I guess we are very close to the final minimal changes 😄

src/lexer.c
```diff
@@ -250,29 +250,65 @@ void tryParseImportStatement () {
...
+        return;
+      }
+
+      while (pos < end) {
+        ch = *pos;
+        if (isQuote(ch)) {
+          stringLiteral(ch);
+        } else if (ch == '}') {
+          pos++;
+          break ;
```
guybedford, line 291:
```suggestion
          break;
```

src/lexer.c
```diff
@@ -250,29 +250,65 @@ void tryParseImportStatement () {
...
+      }
+
+      ch = commentWhitespace(true);
+      if (!str_eq4(pos, 'f', 'r', 'o', 'm')) {
+        syntaxError();
+        break ;
+      }
+
+      pos += 4;
+      ch = commentWhitespace(true);
```
guybedford, line 304: Perfect.

src/lexer.c
```diff
@@ -250,29 +250,65 @@ void tryParseImportStatement () {
...
+          break ;
+        }
+
+        pos++;
+      }
+
+      ch = commentWhitespace(true);
+      if (!str_eq4(pos, 'f', 'r', 'o', 'm')) {
+        syntaxError();
+        break ;
```
guybedford, line 300:
```suggestion
        break;
```

src/lexer.c
```diff
@@ -250,29 +250,65 @@ void tryParseImportStatement () {
...
+      }
+
+      pos += 4;
+      ch = commentWhitespace(true);
+
+      if (isQuote(ch)) {
+        readImportString(startPos, ch);
+      }
```
guybedford, line 308: Strictly speaking we can throw on if we hit `!isQuote`.

src/lexer.c
```diff
@@ -403,14 +471,34 @@ void tryParseExportStatement () {
...
+    // export { mod as "spa ce" } from
+    // export { mod as " space" } from
+    // export { mod as "space " } from
+    // export { mod as "not~id" } from
+    // export { mod as "%notid" } from
+    else {
+      startPos = pos + 1;
+
+      stringLiteral(ch);
+      readToWsOrPunctuator(ch);
```
guybedford, line 495: This line / call isn't needed - just string literal gets us to the right position.

src/lexer.c
```diff
@@ -361,10 +397,42 @@ void tryParseExportStatement () {
...
+        // export { "%notid" as } from
+        // export { "identifer" } from
+        // export { "%notid" } from
+        else {
+          char16_t *posQuoteStart = pos;
+
+          stringLiteral(ch);
+
+          char16_t *posQuoteEnd = pos;
+          readToWsOrPunctuator(ch);
```
guybedford, line 422: This line isn't necessary

src/lexer.c
```diff
@@ -361,10 +397,42 @@ void tryParseExportStatement () {
...
+          ch = commentWhitespace(true);
+
+          if (str_eq2(pos, 'a', 's')) {
+            startPos = pos;
+            endPos = pos;
+          } else {
+            startPos = posQuoteStart + 1;
+            endPos = posQuoteEnd;
+          }
```
guybedford, line 431: This entire branch shouldn't be necessary once you remove the line above.

guybedford (commented):
Still nearly there!

src/lexer.c
richardo2016: Not really, this logic is irrelevant to `readToWsOrPunctuator(ch);` above:

**cause**: we must ensure `readExportAs` always deals with one token with pattern `[as] <"@unicode"|identifer>`.
**result**: to ensure that, in export statement, if we met quote first, we must predetect if there are `as` keyword to determine args `startPos`/`endPos` passed to `readExportAs`

For instance, `export { "~123" as foo0 } from './mod0.js';`:

1. met quote `"`, record `posQuoteStart`, do `stringLiteral`, record `posQuoteEnd`.
2. now we don't if `[as] <"@unicode"|identifer>` followed after where `posQuoteEnd` represents, we so need to detect possible `as` keyword.
    - if `as` detected, adjust `startPos` to `pos`;
    - if `as` not detected, this statment is expected like `export { "~123" } from './mod0.js';`. So we adjust `startPos = posQuoteStart + 1;` & `endPos = posQuoteEnd - 1;` to make `readExportAs` know the truly start position of module name.

So I kept this logic in latest commit.

guybedford: This reminds me we still need to consider the string decoding cases of ensuring that `"\a"` is decoded into "a" etc. We have decoding algorithms in `src/lexer.js` but they only engage for `d === -1`. We probably want to have the end and start match up to the quotes and then have a new side channel for determining encoding being set.

I will put some time aside soon to try and implement this scenario on top of this PR logic. I think it's 90% of the way at this point though.

richardo2016: Your remind me too, for `export { foo0 as ' slash\\ ' } from './mod0.js';`, we expect to got result export-string `' slash\ '`, so should we make corresponding `Export` C-struct like below?

```
// `start` represents space after start quote, and `end` represents space before end quote.
// `quote = 1` means single wrapped by single quote, `2` means double quote, 0 means no quote
{ start: 19, end: 28, quote: 1 }
```

and we can add C-method `eq`(means `getExportQuote`) to check if this string wrapped by quotes.

guybedford: Actually there's no need for a separate method - we should just make the string include the quotes then in `src/lexer.js` do a check for the quotes there to parse them. This is fine because a quote is an invalid character to start an identifier with, so that's all the information we need.

The check would then be `source.charCodeAt(start) === "'" || '\''`

richardo2016: That's also OK, I have another revison do like so, so should I just push it here?

guybedford: Yes definitely

richardo2016: OK I pushed the revision, I found it we just keep quotes in C codes and do escape work in `./src/lexer.js`, our modification would be so simple! @guybedford

src/lexer.js
```diff
@@ -25,7 +25,16 @@ export function parse (source, name = '@') {
...
+  while (wasm.re()) {
+    const s = wasm.es(), e = wasm.ee();
+    const ch = source[s];
+    
+    if (ch === '"' || ch === '\'') {
+      exports.push(decode(source.slice(s, e)));
+    } else {
+      exports.push(source.slice(s, e));
+    }
+  }
```
guybedford, line 37: Since file size is so important, we can save a few bytes here (assuming I've got it right!):

```suggestion
  while (wasm.re()) {
    const expt = source.slice(wasm.es(), wasm.ee()), ch = expt[0];
    exports.push((ch === '"' || ch === "'") ? decode(expt) : expt);
  }
```

guybedford (approved):
Yes, that's it, thanks for the persistence here!

PR #94, opened 2021-10-08
feat: support ci on actions.

richardo2016:
This PR add support to CI based on github actions. Once new commits pushed to specified branchs(e.g, `main`), or any PR created to original repository's `main` branch, actions would be triggered to do those check on **ALL** OS (windows/macOS/Linux):

- if compilation to WASM valid
- if lexer (wasm version) valid
- if compilation to ASM.js valid
- if lexer (asm.js version) valid
- benchmark

You can see process of actions [here](https://github.com/richardo2016-forks/es-module-lexer/actions/runs/1319063400), I have tested it several times @guybedford 

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,158 @@
+name: Test es-module-lexer
+
+on:
+  push:
+    # Sequence of patterns matched against refs/heads
+    branches:
+      - 'feat/**'
+      - 'fix/**'
+      - 'main'
+  pull_request:
+    branches:
+      - 'main'
+
+env:
+  WASI_VERSION: 12
+  WASI_VERSION_FULL: "12.0"
+  WABT_VERSION: "1.0.24"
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,158 @@
+name: Test es-module-lexer
+
+on:
+  push:
+    # Sequence of patterns matched against refs/heads
+    branches:
+      - 'feat/**'
+      - 'fix/**'
+      - 'main'
+  pull_request:
```
richardo2016, line 10: Github actions runs not only on this repository, but also should run on the forkd one which PR was taken from. Developers could make one branch for PR, which has name started with `feat/` for new feature or `fix/` for some fixup, just like what I do this time. 

So while rule for pushing/PR-to `main` branch was kept, I still leave two special rules :

- `feat/**`
- `fix/**`

@guybedford

Issue #127, opened 2022-08-03
Gauging interest in upstreaming some changes


aomarks:
I'm currently integrating es-module-lexer into a project within Google, as part of integrating https://github.com/google/playground-elements/, and in doing that there are a few changes I'm considering.

I'm wondering if you have any thoughts on these, and whether you'd be interested in receiving upstream PRs for them?

1. Converting the `lexer.js` bridge to TypeScript. This was really straightforward since I was able to just integrate the typings already available in `lexer.d.ts`.

2. Replace the use of `eval` for JavaScript string literal unescaping with a C version. `eval` is a security smell for obvious reasons, so even though the implementation here is likely safe because it's only passed strings that have already been broadly parsed as strings, we have some auditing that makes it difficult to use `eval` at all. Also it requires broadening CSP. So my thought was to extend the C lexer to deal with unescaping directly. Curious if you already thought about that, whether `eval` turned out to be faster, or just simpler/less code?

3. I'm curious about the manual heap management at https://github.com/guybedford/es-module-lexer/blob/a28c3668119b538e18c459baf14e630bc27ff118/src/lexer.js#L10, and was wondering if using `malloc` in the C program would be a simpler/safer approach? Maybe it was that the malloc implementation from wasi inflated the code too much? Curious to hear your thoughts on that.

Thanks for the awesome library! Would like to contribute if any of these changes sound good. Otherwise we can just maintain a small internal fork.

aomarks:
Oh and one more thought:

4. `parse` currently either returns a result array or the promise of one, depending on whether initialization has completed. What would you think about `init` instead resolving to the `parse` function, not exporting `parse` directly? That way the user can't accidentally write racey code where sometimes unexpectedly get a promise? Plus it slightly simplifies the `parse` implementation, since it could assume `wasm` is always defined?

So usage would look like:

```ts
import * as esmLexer;
const parse = await esmLexer.init;
const [imports, exports] = parse(...);
```

guybedford:
Hey @aomarks thanks for showing an interested in contributing and for providing such a clear contribution plan! That's a great approach to collaborating.

Would be very much open to new directions, although two of the major goals for this project are performance and size. In particular, this project was originally designed for es-module-shims. As a result, a heavy emphasis is placed on new features to entirely justify their overhead in terms of footprint, so we just need to work within these constraints.

Some feedback on your points:

[...]

PR #130, opened 2022-08-10
Convert WASM bridge to TypeScript

aomarks:
As discussed in https://github.com/guybedford/es-module-lexer/issues/127, this PR converts the WASM JavaScript bridge to TypeScript.

I suggest looking at each commit individually for the best diffs. There's an intermediate commit that shows the differences between the original and newly generated `.js` and `d.ts` files. The next commit deletes them, since I don't think we want those artifacts checked in?

There shouldn't be any user-affecting changes in this PR. `dist/lexer.js` ends up being byte-for-byte identical, and `types/lexer.d.ts` should be semantically identical (verified manually), though with slightly different formatting.

For the build, I am using swc + chomp to generate the `.js` file as requested in https://github.com/guybedford/es-module-lexer/issues/127#issuecomment-1204629293. It worked great! However, since swc [doesn't generate `.d.ts` files](https://github.com/swc-project/swc/issues/657), I also included a chomp `tsc` step with the `--emitDeclarationOnly` flag. Hopefully this is OK? The two steps can at least run in parallel.

I did hit a small issue with the chomp swc template: it seems like the `source-maps = false` option doesn't take effect -- I am still seeing a `lexer.js.map` file being generated. Any idea what might be going on there?

Thanks!


src/lexer.ts
```diff
@@ -137,24 +138,129 @@ export interface ExportSpecifier {
...
+    const n = source.slice(s, e), ch = n[0];
+    const ln = ls < 0 ? undefined : source.slice(ls, le), lch = ln ? ln[0] : '';
+    exports.push({
+      s, e, ls, le,
+      n: (ch === '"' || ch === "'") ? decode(n) : n,
+      ln: (lch === '"' || lch === "'") ? decode(ln) : ln,
+    });
+  }
+
+  function decode (str: string|undefined) {
```
guybedford, line 193: Hate to nitpick, but may as well since it's the only case:

```suggestion
  function decode (str: string | undefined) {
```

chompfile.toml
```diff
@@ -43,6 +43,26 @@ target = 'dist/lexer.cjs'
...
 run = 'babel dist/lexer.js | terser -c -m -o dist/lexer.cjs'
 
+[[task]]
+name = 'build:swc'
+target = 'src/lexer.js'
+dep = 'src/lexer.ts'
+template = 'swc'
+[task.template-options]
+# TODO This still seems to generate lexer.js.map
+source-maps = false
```
guybedford, line 53: Seems this is an SWC issue - https://github.com/swc-project/cli/issues/113.

One easy fix might be to eject the SWC plugin usage (`chomp --eject`) can be used to see what this looks like, Ideally we could eject specific tasks but that isn't a feature currently.

Then the SWC task could be updated to:

```toml
run = '''
./node_modules/@swc/cli/bin/swc ....
rm src/lexer.ts.map
'''
```

guybedford (approved):
Thanks for posting, this all looks great to me, happy to merge and release, just see the two comments in the review.

chompfile.toml
aomarks: Ah ha, thanks for finding that! Applied the ejected output just for this task, with with a note that we should switch back to the swc template once that issue is fixed.

aomarks:
> Thanks for posting, this all looks great to me, happy to merge and release, just see the two comments in the review.

Thanks! Comments addressed.

chompfile.toml
```diff
@@ -43,6 +43,27 @@ target = 'dist/lexer.cjs'
...
 
+[[task]]
+name = 'build:swc'
+target = 'src/lexer.js'
+dep = 'src/lexer.ts'
+# Note we should use the chomp swc template, but
+# https://github.com/swc-project/cli/issues/113 means we always get a sourcemap
+# even when we set "source-maps = false", so for now we have ejected the
+# template to its raw "run" command, and added an "rm" step.
+run = 'node ./node_modules/@swc/cli/bin/swc.js $DEP -o $TARGET --no-swcrc -C jsc.parser.syntax=typescript -C jsc.parser.importAssertions=true -C jsc.parser.topLevelAwait=true -C jsc.parser.importMeta=true -C jsc.parser.privateMethod=true -C jsc.parser.dynamicImport=true -C jsc.target=es2016 -C jsc.experimental.keepImportAssertions=true && rm src/lexer.js.map'
```
guybedford, line 54: One of the main reasons I worked on Chomp is because I run windows and got really annoyed with the lack of powershell support. Unfortunately it doesn't support `&&` and `||` just yet, so would be great to stick to multiline for now:

```suggestion
run = '''
node ./node_modules/@swc/cli/bin/swc.js $DEP -o $TARGET --no-swcrc -C jsc.parser.syntax=typescript -C jsc.parser.importAssertions=true -C jsc.parser.topLevelAwait=true -C jsc.parser.importMeta=true -C jsc.parser.privateMethod=true -C jsc.parser.dynamicImport=true -C jsc.target=es2016 -C jsc.experimental.keepImportAssertions=true
rm src/lexer.js.map
'''
```

PR #135, opened 2022-08-30
feat: support local names for export default

### LINKED ISSUE #125: Local names on exports for export default declaration forms
`export default (async)? function f` and `export default class x` should set local name to the name of the function or class.

await-ovo:
To Resolve #125


src/lexer.c
```diff
@@ -418,10 +420,45 @@ void tryParseExportStatement () {
...
-      case 'd':
-        addExport(pos, pos + 7, NULL, NULL);
+      case 'd': {
+        const char16_t* startPos = pos;
+        pos += 7;
+        ch = commentWhitespace(true);
+        // export default async? function*? name? (){}
+        if (ch == 'a' && memcmp(pos + 1, &SYNC[0], 4 * 2) == 0) {
+          pos += 5;
+          ch = commentWhitespace(true);
```
guybedford, line 430: I believe `async` and `function` need to be on the same line so that this should be `commentWhitespace(false)` strictly speaking. 

And this does actually have a real case it handles though believe it or not:

```js
const async = 2;
export default async

function x () {
}
```

Then in such a case, we need to ensure the fallthrough is handled properly as well instead of checking the next if statement.

guybedford (commented):
Thanks for the PR, see edge case discussed - we need to be careful about the if statement fallthroughs as well.

src/lexer.c
await-ovo: Thanks for the review and the great case, aha, I just learned that `async` is not a reserved word, I also handled the following case:
```
const asyncVar = 1
export default asyncVar
```

PR #160, opened 2023-11-09
feat: ESM detection support

### LINKED ISSUE #159: Feature request: Include an information whether the code includes `export {}`
Currently `es-module-lexer` doesn't return any information that `export {}` existed.
https://stackblitz.com/edit/node-1bnjp9?file=package.json,index.js

It would be useful if `es-module-lexer` could return that information. Because `export {}` is sometimes used to indicate that the file is written in ESM.

I'm not sure how to put this information in the current API. Maybe returning -1 for `ExportSpecifier['s']`?

sapphi-red:
close #159

I added the fourth return value that returns whether the code includes ESM syntaxes.

I didn't change https://github.com/guybedford/es-module-lexer/blob/main/lexer.js. Is this file used?


guybedford (approved):

guybedford:
`lexer.js` is an old browser build that is no longer maintained, although the tests can be run manually. It should probably just be removed at this point.

PR #162, opened 2023-11-09
feat: skip trailing white space in dynamic import

### LINKED ISSUE #118: Parse dynamic imports with backticks
When parsing a dynamic import with backticks, the `s` and `e` result can be used to get the raw string. However the `e` result seems to always be one index less of the closing parenthesis/comma, instead of the last backtick. 

# Reproduction

[Stackblitz](https://stackblitz.com/edit/node-gjj4bq?file=index.js)

Run `node index.js` in the terminal to log the first parsed import

# Expected result

The log should be ```"`bla.js`"```

# Actual result

It gives ```"`bla.js`\n  /* bar */\n"``` (trailing comments aren't stripped)

sapphi-red:
close #118

I'm not sure if this is always correct.


guybedford:
I'm not sure this works either, it could be worth adding some edge cases for various parse goals to verify further.

sapphi-red:
I added some more test cases that I can think of.

guybedford:
Nice work, I can't think of any edge cases actually. Will release and let's see if anything comes up further.

PR #164, opened 2023-11-10
fix: ignore dynamic import for `hasModulesyntax`

sapphi-red:
I initially included dynamic imports to be detected as `hasModuleSyntax = true` in #160.
But now I recalled that non-ESM files (e.g. CJS files, `<script>` tag) can contain dynamic imports.

This PR will make dynamic import to be ignored for `hasModulesyntax`. Including a dynamic import won't make `hasModulesyntax` to be `true` anymore.

Users that wants the previous behavior can achieve with `imports.some(i => i.d >= 0) || hasModuleSyntax`.


guybedford:
Thanks for catching this and posting the follow-up.

PR #179, opened 2024-07-09
fix: add exports type for lexer.asm.js

ahaoboy:


package.json
```diff
@@ -12,7 +12,11 @@
       "import": "./dist/lexer.js",
       "require": "./dist/lexer.cjs"
     },
-    "./js": "./dist/lexer.asm.js"
+    "./js": {
+      "types": "./types/lexer.d.ts",
+      "module": "./dist/lexer.asm.js",
+      "import": "./dist/lexer.asm.js"
```
guybedford, line 18: This can just be `"default"` since we don't have any other option:

```suggestion
      "default": "./dist/lexer.asm.js"
```

PR #183, opened 2024-12-07
Add side effects to package.json

denisx:
Add side effects to package.json, help bundlers cut unused code with treeshake. Good for esm, modules, named imports.


BridgeAR (approved):
I guess this does not hurt either way

PR #185, opened 2024-12-14
feat: implement `initSync`

### LINKED ISSUE #184: Init `es-module-lexer` synchronously
So I want to use `es-module-lexer` in an ESlint plugin (https://github.com/un-ts/eslint-plugin-import-x/issues/201).

The ESLint plugin requires rules to be synchronous, so `await init` is really not an option for us. However, I believe it is possible to init wasm synchronously using `WebAssembly.Module` and `WebAssembly.Instance` API (at least in Node.js), see https://github.com/nodejs/cjs-module-lexer/blob/8411abde37566a9d5d1e34fe3c6952faae5badc1/src/lexer.js#L111-L119

SukkaW:
Fixes #184 by implementing `initSync` using `WebAssembly.Module` and `WebAssembly.Instance`.

I have not added a unit test case for `initSync` since I don't know which test file to add. Please give me a review comment on this if you believe a unit test is needed.

### DIFF
--- a/src/lexer.ts
+++ b/src/lexer.ts
@@ -294,13 +294,24 @@ let wasm: {
   ss(): number;
 };
 
+const getWasmBytes = () => (
+  binary => typeof Buffer !== 'undefined'
+    ? Buffer.from(binary, 'base64')
+    : Uint8Array.from(atob(binary), x => x.charCodeAt(0))
+)('WASM_BINARY');
 
 /**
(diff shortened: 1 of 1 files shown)


SukkaW:
@guybedford Just a friendly ping. Would you like to review this PR and share your thoguhts?

guybedford:
Sure I'm happy to land this for Node.js / non-browser use cases.

PR #197, opened 2026-04-20
Always exclude dyn import attributes from specifier range

### LINKED ISSUE #196: Dynamic import specifier position includes attributes when using a non-string specifier
```js
parse(`import("foo", {})`)[0][0]
parse(`import(foo, {})`)[0][0]
```

- The first one gives a `s...e` range covering `"foo"` 
- The second one gives a `s...e` range covering `foo, {}`

I would expect `s...e` to either include import attributes in both cases, or to always exclude them.

nicolo-ribaudo:
Fixes https://github.com/guybedford/es-module-lexer/issues/196


src/lexer.c
```diff
@@ -127,6 +127,22 @@ bool parse () {
...
         break;
+      case '[':
+        openTokenStack[openTokenDepth].token = AnyBracket;
+        openTokenStack[openTokenDepth++].pos = lastTokenPos;
+        break;
+      case ']':
+        if (openTokenDepth == 0)
+          return syntaxError(), false;
+        openTokenDepth--;
+        break;
```
nicolo-ribaudo, line 138: Without this, we'd consider the `,` inside `[...]` inside a dynamic import specifier to end the specifier part (e.g. `import([x,y],z)`, we'd consider the specifier to be only `[x`).

- Do you think it's worth adding this case?
- Should I only consider there two branches when `dynamicImportStackDepth > 0`, or is this fine?

src/lexer.h
```diff
@@ -57,6 +57,7 @@ enum OpenTokenState {
   ImportParen = 5, // import(),
   ClassBrace = 6,
   AsyncParen = 7, // async()
+  AnyBracket = 8, // [
```
nicolo-ribaudo, line 60: There is a `Paren = odd, Brace = even` comment at the beginning of this enum. I don't think it actually matters, looking at the code and tests.

src/lexer.c
guybedford: This works fine!

src/lexer.h
guybedford: Removed the comment about odd/even.

guybedford:
Pushed some extra test cases.

guybedford:
Released in 2.1.0.

PR #199, opened 2026-06-16
fix: detect every binding in export var/let/const declarations

### LINKED ISSUE #175: Support renamed destructured exports
```ts
const { init, parse } = require('es-module-lexer')

init.then(() => {
  const [imports, ex] = parse(`
const o = { name5: 1, name6: 1 }
export const { name5, name6: bar } = o`
  )
  console.log(imports, ex)
})
```
Outputs
```ts
[] [
  { s: 49, e: 54, ls: 49, le: 54, n: 'name5', ln: 'name5' },
  { s: 56, e: 61, ls: 56, le: 61, n: 'name6', ln: 'name6' }
]
```
But there should be a `bar` export?

### LINKED ISSUE #176: Support multiple exports per declaration
```ts
const { init, parse } = require('es-module-lexer')

init.then(() => {
  const [imports, ex] = parse('export const name1 = 1, name2 = 2;')
  console.log(imports, ex)
})
```
Outputs
```ts
[] [ { s: 16, e: 21, ls: 16, le: 21, n: 'name1', ln: 'name1' } ]
```
But there should be a `name2` export?

BridgeAR:
`export var`/`let`/`const` declarations reported only their first binding. The declaration scanner stopped at the first `= initializer`, so every later declarator was dropped (`export const a = 1, b = 2` reported only a), and the destructuring path reported the property key rather than the bound name (`export const { a: b } = x` reported a), missing defaults, nested patterns and rest elements.

The scanner now walks the whole binding list: it reads each bound name and skips every initializer and default value expression-aware, so a comma inside one no longer ends the list, and the list terminates at `;`, EOF or an ASI line break (never reading the next statement as a binding).

That expression skip shares the main tokenizer (a single `consumeToken`), so its regex / keyword / import() resolution cannot diverge from the main scan. 

Fixes #176
Fixes #175


guybedford:
Really nice to finally get every binding name out of these declarations, and the destructuring walk in `readBindingPattern`/`readBindingTarget` is clean — that part is genuinely new work and reads well.

My one substantive concern is `skipExpression`. It re-implements regex-vs-division disambiguation from scratch (`isValueChar` + `lastWasValue`), and that's a weaker copy of logic the lexer already does correctly in the main loop. Because it only looks at the previous *character* rather than the previous *token*, it misses the value-keyword case, and on valid input that now throws rather than under-reporting:

```js
export const f = () => { return /,/g }, b = 2;   // → throws "Parse error"
```

`return` ends in a value char, so the `/` is read as division; the regex's closing `/` then opens a second, phantom regex that runs to EOF. On `main` this parsed fine (it bailed at the first `=`), so this is a regression — and throwing is strictly worse than the old under-reporting, since it makes the whole module un-analyzable.

The deeper point is that `skipExpression` shouldn't own this decision at all. The main tokenizer already tracks everything needed and resolves the ambiguity properly at `src/lexer.c:204-238`, keyed off real lexer state:

- `lastTokenPos` — the previous **token** (so `isExpressionKeyword` catches `return`/`typeof`/`in`/etc., `isExpressionPunctuator` catches operators, `isParenKeyword` catches `if (...) /re/`)
- `openTokenStack` / `openTokenDepth` — nesting, including which token preceded each opening bracket (`ClassBrace`, `isExpressionTerminator`, the `for/of` binding case)
- `lastSlashWasDivision` — the sticky flag for `/` chains

`skipExpression` re-derives a thin version of all of this (its local `depth` shadows `openTokenStack`, its `lastWasValue` shadows the `:204` decision) and also re-dispatches strings/templates/comments that the main loop already routes through `stringLiteral`/`templateString`/`lineComment`/`blockComment`. So we have two divergent copies of a subtle invariant, and the new one is the buggy one.

Two ways to share the real machinery, in order of preference:

**1. Ride the main loop.** The only thing the binding scan genuinely needs that the tokenizer doesn't already give us is "is this `,`/`;` the end of the *binding list*?" — which is just `openTokenDepth == <depth recorded at the declaration>`. If we extract the per-token body of the `mainparse` switch into a `consumeToken()` that updates the globals, then both the main loop and the initializer-skip call it, and the skip just adds the terminator check after each token. Regex, strings, templates, comments, and `lastSlashWasDivision` all stay correct by construction because it *is* the tokenizer.

**2. Minimal: factor just the decision.** Pull `:204-238` into a `bool slashIsRegex()` and have `skipExpression` use the global `openTokenStack`/`openTokenDepth`/`lastTokenPos` instead of its private `depth`/`lastWasValue`. Smaller diff, kills the divergent regex copy and the throw, but `skipExpression` still re-dispatches strings/templates/comments — so less clean than (1).

Either way the `return /,/g` case goes away for free, and we stop maintaining two copies of the regex heuristic.

The genuinely new logic that has to stay in the export path regardless: ASI termination (line break after a value), since the main loop doesn't care about statement boundaries, and the binding-target/pattern reading. Both are fine as-is.

BridgeAR:
I updated the code to use as much of the main loop as possible. I tried to make sure the performance stays the same while being able to parse more and that was the reason I pulled out `consumeToken`.

guybedford (approved):

PR #202, opened 2026-06-21
build: replace emscripten-fastcomp asm.js build with binaryen wasm2js

BridgeAR:
## Summary

The `es-module-lexer/js` build was compiled by emscripten 1.40.1-fastcomp, a frozen toolchain with no arm64 distribution that drops the wasm data section. The asm.js wrapper therefore carried a hand-maintained copy of `lexer.c`'s keyword constants at a fixed memory offset, so any change to `parse()`'s codegen silently desynced that dictionary, and fastcomp miscompiled a restructured main loop outright.

`wasm2js` (binaryen) compiles the existing `lexer.wasm` to JS with the data section embedded, so the wrapper drives the module like the wasm build (`memory.grow` + `sa` + `parse`) and there is no dictionary to keep in sync. CI installs `binaryen` via apt instead of cloning emsdk; the committed `lib/` artifacts are regenerated with wasi-sdk 12 + binaryen.

## Test plan

- [ ] `chomp test` green on both the wasm and `es-module-lexer/js` builds


guybedford:
This is great, and a really nice restructure, but unfortunately there seems to be a roughly 30% performance drop with using this approach. Here are my results:

| build | asm.js warm | wasm warm | asm.js cold |
|---|---|---|---|
| emscripten-fastcomp | 16.64ms | 13.12ms | 32ms |
| wasm2js baseline | 21.32ms | 13.40ms | 95ms |
| wasm2js + `-O3` + inline memcmp | 21.92ms | 12.96ms | 101ms |

Therefore I think it likely cannot land unfortunately. In general, manually updating the dictionary is indeed fiddly but also relatively straightforward, it just requires some messing with the offsets.

guybedford:
I've posted an alternative in https://github.com/guybedford/es-module-lexer/pull/203.

BridgeAR:
I noticed the performance impact of the default build as well and was still working on that 😅 

I managed to actually make it faster than using fastcomp in the end, so my suggestion is to still change the toolchain due to the age of fastcomp and it not being developed anymore.

The reason I started working on this was that fastcomp could not compile my changes needed to make the new code fast. I did not yet check if that is now resolved or not.

BridgeAR:
I made it work with the current tooling and that does still seem to outperform the new tooling. Closing

PR #210, opened 2026-06-27
fix: don't report a method named import as a dynamic import

### LINKED ISSUE #209: 2.1.0: a method named `import` is reported as a dynamic import (regression vs 2.0.0)
### Summary

In `es-module-lexer@2.1.0`, a **class/object method named `import`** is reported as a **dynamic import** (`import(...)`). `2.0.0` and `1.7.0` do not exhibit this — so this is a regression introduced in 2.1.0.

This breaks any tooling that rewrites detected dynamic imports, because the lexer points at a method-parameter list rather than a real `import()` call.

### Minimal reproduction

```js
import { init, parse } from 'es-module-lexer';
await init;

const code = `class Graph {\n\timport(data, merge = false) {\n\t\treturn this;\n\t}\n}`;
console.log(parse(code)[0]);
```

Observed:

| es-module-lexer | result for the `import(...)` **method** |
| --- | --- |
| `1.7.0` | `[]` (correct — not an import) |
| `2.0.0` | `[]` (correct — not an import) |
| **`2.1.0`** | **`[{ d: 21, n: undefined, … }]`** (WRONG — flagged as a dynamic import) |

A genuine dynamic import (`const x = await import('graphology');`) is still detected correctly (`n: 'graphology'`).

### Expected

A method named `import` (class method or object shorthand) must **not** be classified as a dynamic import — restore the 1.7.0 / 2.0.0 behavior.

### Downstream impact

Vite **8.1.0** bundles es-module-lexer **2.1.0**. Its import-analysis sees the false-positive dynamic import and wraps the method's first parameter with its query-injection helper, producing invalid JS like:

```js
import(__vite__injectQuery(data, 'import'), merge = false) { … }
```

which throws `SyntaxError: Unexpected token '('` at module-evaluation time. This is hit in practice by any library exposing an `import` method — e.g. **graphology** (`Graph.prototype.import(data, merge)`), which fails to load under Vite 8.1.0 in browser/test contexts.

### Environment

- es-module-lexer 2.1.0 (regression vs 2.0.0 / 1.7.0)
- Surfaced via vite 8.1.0 (bundles 2.1.0) + graphology in a SvelteKit + Vitest browser-mode setup.

BridgeAR:
## Summary

A class or object method named `import` whose parameter list contains a comma, such as `class Graph { import(data, merge = false) {} }`, was reported as a dynamic `import()`. Vite's import-analysis then rewrote the method's first parameter into invalid JS, so any library exposing an `import` method (e.g. graphology's `Graph.prototype.import`) failed to load under Vite 8.1.0.

`import(` always provisionally registers a dynamic import; the `) {` heuristic later removes it when the call turns out to be a method body. That removal matched `import->end` against the closing `)`, which only held by coincidence when the last argument ended adjacent to the `)`. A comma in the argument list moves `end` to the end of the first argument, so the removal stopped firing and the false positive leaked out. The fix matches against `statement_end` (`se` in the JS port), which is the `)` position for both `import()` and `import(a, b)`.

Regression introduced in 2.1.0 by the import-attribute comma handling (#197).

## Test plan

- [ ] `chomp test` (wasm + asm) green; both new cases pass on the wasm and asm builds
- [ ] Real dynamic imports with a comma argument (the `Dynamic import expression range 7-17` and attribute cases) still detected

Fixes: https://github.com/guybedford/es-module-lexer/issues/209


BridgeAR:
I just saw the duplicate

PR #212, opened 2026-06-29
docs: correct stale limitations and gzip size in README

BridgeAR:
## Summary

Two README claims no longer match the shipped library:

1. The **Limitations** section states that `export var a = 'asdf', q = z` only
   detects `a` (not `q`) and that `export var { a: b } = asdf` misses `b`. Both
   are detected correctly in current releases (verified against 2.2.0):

   ```js
   parse("export var a = 'asdf', q = z;")  // exports: ["a", "q"]
   parse("export var { a: b } = asdf;")    // exports: ["b"]
   ```

   The examples are replaced with the limitation that actually holds: the lexer
   does not validate, so some invalid source (e.g. `export const = 1`) is
   accepted without an error rather than rejected. This is already implied under
   "Grammar Support" ("may parse invalid JS source without errors"); the
   Limitations section now states it directly.

2. The **"4KiB gzipped"** figure is about half the current size. All three
   published builds gzip to ~8 KiB:

   | file | raw | gzip |
   | --- | --- | --- |
   | `dist/lexer.js` | 17132 | 7916 |
   | `dist/lexer.cjs` | 17211 | 7922 |
   | `dist/lexer.asm.js` | 27338 | 7741 |

No code changes; docs only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Outputs the list of exports and locations of import specifiers, including dynami
 
 Supports new syntax features including import attributes and source phase imports.
 
-A very small single JS file (~8KiB gzipped) that includes inlined Web Assembly for very fast source analysis of ECMAScript module syntax only.
+A very small single JS file (~7KiB gzipped) that includes inlined Web Assembly for very fast source analysis of ECMAScript module syntax only.
 
 For an example of the performance, Angular 1 (720KiB) is fully parsed in 5ms, in comparison to the fastest JS parser, Acorn which takes over 100ms.
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Outputs the list of exports and locations of import specifiers, including dynami
 
 Supports new syntax features including import attributes and source phase imports.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,7 @@ Outputs the list of exports and locations of import specifiers, including dynami
 
 Supports new syntax features including import attributes and source phase imports.
 
-A very small single JS file (4KiB gzipped) that includes inlined Web Assembly for very fast source analysis of ECMAScript module syntax only.
+A very small single JS file (~8KiB gzipped) that includes inlined Web Assembly for very fast source analysis of ECMAScript module syntax only.
```
guybedford: I usually report brotli sizes here which gives 7KiB.

guybedford:
```suggestion
A very small single JS file (~7KiB gzipped) that includes inlined Web Assembly for very fast source analysis of ECMAScript module syntax only.
```

guybedford (approved):

Issue #72, opened 2021-06-07
Can not parse `export type {} from 'xxx'`


qmhc:
When I use:

```ts
import type { Test } from './test'

export type { Test }
```

it can be parsed, but not:

```ts
export type { Test } from './test'
```

guybedford:
The lexer is only designed to support JavaScript not TypeScript. If you need TypeScript support that should probably be implemented as a fork of this project as I don't want to deviate the design goals here towards that currently.

qmhc:
Thanks for your reply, but I am not good at c lang, maybe I will try in the future.

PR #213, opened 2026-06-29
feat: lex type-only TypeScript imports and exports

BridgeAR:
## Summary

Teaches the lexer to understand the type-only TypeScript syntax that fails or mis-parses today (issue #72: `export type { A } from 'x'`), without a separate transform step and without pulling a TypeScript parser into the package.

The existing C lexer learns the syntax behind a `LEX_TS` compile-time switch that produces a third Wasm build alongside the full and minimal ones:

1. The JavaScript build is the same code path with no added branches — non-TypeScript users pay nothing. The TypeScript Wasm is a few hundred bytes larger than the JavaScript one and ships only behind a new `es-module-lexer/ts` entry exposing `parseTs` / `initTs`.
2. Type-only imports and exports are reported rather than elided, marked with a new `tp` boolean on every import and export specifier (always `false` from `parse`). Inline modifiers are tracked per specifier, so `export { type A, b }` marks only `A`.

## Why

Running a TypeScript transform first is no faster than what Node.js type stripping already does at runtime; a single lexer that handles both languages is the win. Reporting (rather than eliding) type-only specifiers serves consumers building a module graph, who need to see the type edges.

## Notes for review

- Stacked on #211 (the minimal build): review this PR's diff against the `lexer-min` base, not `main`.
- The minimal build lexes JavaScript only and never sets `tp`; its reader and generated `.d.ts` drop the field, so es-module-shims' footprint is unaffected. Only the full and TypeScript builds carry `tp`.
- The TypeScript reader (`src/lexer-ts.ts`) reports export `ss` to match the full `parse` shape now that #206 has landed on this base.
- `lib/lexer.ts.wasm` and the other build artifacts are produced by `chomp build`; the committed tree carries source only, leaving the binaries to the canonical toolchain / CI.

## Scope

The accept boundary matches Node.js type stripping. This first increment covers type-only imports and exports. Type annotations, generics, `as` / `satisfies`, non-null `!` and the rest of the erasable surface are planned as a follow-up; non-erasable syntax (`enum`, runtime `namespace`, parameter properties, legacy decorators) stays out of scope.

## Test plan

- `chomp test` runs the new `test:ts` job (TS=1) alongside the existing full and minimal wasm/asm suites.
- New suites under `test/typescript/`: type-only imports, type-only exports, regex/division non-regression, attribute/phase permutations, and a JS-superset guard asserting plain-JS inputs report `tp: false`.

Fixes: #72


guybedford:
I would be happy to include this by default (or behind a switch / parse option) in the non-minimal build, which was kind of my hope with setting up the minimal build in the first place for when we got to this topic!

Please let's keep the two builds only, then I can review this weekend.

guybedford:
I've landed the min build now to lay the groundwork for this, didn't mean to close it, but deleting the branch did that. Please repost.

PR #216, opened 2026-07-01
feat: add CommonJS entry for es-module-lexer/js

BridgeAR:
## Summary

The `es-module-lexer/js` (CSP-safe asm.js) build shipped ESM only, so `require('es-module-lexer/js')` failed. The obvious fix — transpiling the ESM output to CJS with babel, as the Wasm build does — would duplicate the ~22KB asm.js payload into a second file.

Instead the payload is emitted once as a side-effecting script (`dist/lexer.asm.core.cjs`) that publishes `parse` on a shared realm-global keyed by `Symbol.for('es-module-lexer/js')`. The ESM and CJS entry points are thin wrappers (committed under `src/`) that load the core for its side effect and re-expose the symbol. The core has no import/export, so the same bytes are valid under both module systems; the `.cjs` extension makes `require` treat it as CommonJS.

## Why

Adding `require` support should not double the on-disk footprint of the CSP build. The wrapper indirection keeps the payload single-copy: both entry points share one `parse` (same function identity), so there is no size regression versus the ESM-only status quo.

## Test plan

- [ ] `chomp build` produces `dist/lexer.asm.core.cjs` plus the two wrappers, and the wrappers stay tiny (no payload).
- [ ] `chomp test` (asm + wasm suites) pass.
- [ ] `import { parse } from 'es-module-lexer/js'` and `require('es-module-lexer/js')` both expose the same `parse`.

### DIFF
--- a/src/lexer.asm.js
+++ b/src/lexer.asm.js
@@ -20,7 +20,7 @@ const copy = new Uint8Array(new Uint16Array([1]).buffer)[0] === 1 ? function (sr
 const words = '{{WORDS}}';
 
 let source, name;
-export function parse (_source, _name = '@') {
+function parse (_source, _name = '@') {
   source = _source;
   name = _name;
   // 2 bytes per string code point
@@ -229,5 +229,12 @@ function isBr (c) {
 function syntaxError () {
   throw Object.assign(new Error(`Parse error ${name}:${source.slice(0, acornPos).split('\n').length}:${acornPos - source.lastIndexOf('\n', acornPos - 1)}`), { idx: acornPos });
 }
+
+// This file is emitted as a single side-effecting script (dist/lexer.asm.core.cjs)
+// that is loaded by both the ESM and CJS `es-module-lexer/js` wrappers, so the
+// heavy asm.js payload only ships once. Instead of exporting, publish `parse`
+// on a shared realm-global keyed by a well-known symbol; the wrappers read it
+// back and re-expose it under their respective module systems.
+(typeof globalThis !== 'undefined' ? globalThis : self)[Symbol.for('es-module-lexer/js')] = { parse };
 
 // function asmInit () { ... } from lib/lexer.asm.js is concatenated at the end here
--- a/README.md
+++ b/README.md
@@ -172,6 +172,14 @@ For a version that works with CSP eval disabled, use the `es-module-lexer/js` bu
 import { parse } from 'es-module-lexer/js';
 ```
 
+CommonJS is also supported:
+
+```js
+const { parse } = require('es-module-lexer/js');
+```
+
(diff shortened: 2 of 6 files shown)


BridgeAR:
I use this pattern to prevent code duplication between CJS and ESM. I personally like it a lot, while I am also fine to add a regular duplication instead.

guybedford:
Why is this needed in CommonJS environments? Is require(esm) not enough?

BridgeAR:
That is not available in older Node.js versions that we still support

guybedford:
Which versions does import in the middle need to support?

guybedford:
What about creating a es-module-lexer-cjs, that just does

```js
exports.init = function init () {
  ({ exports: { parse } } = await import('es-module-lexer'));
}
```

sort of thing?

BridgeAR:
It has to be true sync and we need Node.js 18 support

src/lexer.asm.js
```diff
@@ -230,4 +230,11 @@ function syntaxError () {
   throw Object.assign(new Error(`Parse error ${name}:${source.slice(0, acornPos).split('\n').length}:${acornPos - source.lastIndexOf('\n', acornPos - 1)}`), { idx: acornPos });
 }
 
+// This file is emitted as a single side-effecting script (dist/lexer.asm.core.cjs)
+// that is loaded by both the ESM and CJS `es-module-lexer/js` wrappers, so the
+// heavy asm.js payload only ships once. Instead of exporting, publish `parse`
+// on a shared realm-global keyed by a well-known symbol; the wrappers read it
+// back and re-expose it under their respective module systems.
+(typeof globalThis !== 'undefined' ? globalThis : self)[Symbol.for('es-module-lexer/js')] = { parse };
```
guybedford, line 238: We can't do a global side-effect unfortunately. I think duplicating the build may be the better option.

Could you rather just have a local build step that does source replacement on the ESM version? `esmSource.replace(/export /, 'module.exports = ')` sort of thing?

BridgeAR:
I did that for now. What is bad about the side effect though? I am curious :)
I tested it against any build system I know and they all work

guybedford:
I'd recommend you just continue to this. I'm happy to provide the guarantee that the build can work under that source replacement if you want to say post a test for that?

> What is bad about the side effect though? I am curious :)

Two things - (1) deoptimizing the global object, and (2) exposing a public Symbol on it breaking encapsulation...

BridgeAR:
I would delete the symbol normally to prevent exposing anything on it. I did not verify the current and latter type

guybedford:
I'd really prefer not to land anything here if possible, short of a test that a replacement pattern can work.

BridgeAR:
Closing for now

PR #223, opened 2026-07-07
perf(lexer): gate keywordStart on the next byte for i/c tokens

BridgeAR:
## Summary

The main character loop calls `keywordStart()` (and through it a second boundary helper) on every `i` and `c` token it reaches. On the angular/d3/rollup/magic-string corpus the byte after the token rules it out ~98% of the time for `i` and ~95% for `c`, so those calls are almost always wasted. Checking `*(pos + 1)` against the keyword's second letter (`m` for `import`, `l` for `class`) first short-circuits the calls with a single load+compare. `memcmp` already requires that same byte, so the guard is behaviour-preserving.

## Why

The per-character loop is the dominant cost (it runs once per source code unit; the boundary helpers are called <0.05x per code unit), so cutting a function call out of the two hottest keyword cases moves the needle where micro-tuning the helpers themselves does not.

Measured on the same corpus the README benchmarks (~3M code units, Node 24 / V8, 150 iters x 13 trials, drop best+worst, run against the built wasm):

- parse time: **-1.4% mean** (every trial negative), parse output byte-identical
- wasm: **-229 bytes / -129 instructions**

## Drive-by

- `isBrOrWs`: write the `9..13` range as `(char16_t)(c - 9) < 5`, one unsigned compare instead of two. The wasm backend keeps the shorter form at every inlined copy of the helper; this accounts for most of the byte/instruction reduction above.

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -84,11 +84,11 @@ static inline __attribute__((always_inline)) bool consumeToken (char16_t ch) {
         tryParseExportStatement();
       break;
     case 'i':
-      if (keywordStart(pos) && memcmp(pos + 1, &MPORT[0], 5 * 2) == 0)
+      if (*(pos + 1) == 'm' && keywordStart(pos) && memcmp(pos + 1, &MPORT[0], 5 * 2) == 0)
         tryParseImportStatement();
       break;
     case 'c':
-      if (keywordStart(pos) && memcmp(pos + 1, &LASS[0], 4 * 2) == 0 && isBrOrWs(*(pos + 5)))
+      if (*(pos + 1) == 'l' && keywordStart(pos) && memcmp(pos + 1, &LASS[0], 4 * 2) == 0 && isBrOrWs(*(pos + 5)))
         nextBraceIsClass = true;
       break;
     case '(':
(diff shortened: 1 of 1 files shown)


src/lexer.c
```diff
@@ -84,11 +84,11 @@ static inline __attribute__((always_inline)) bool consumeToken (char16_t ch) {
         tryParseExportStatement();
       break;
     case 'i':
-      if (keywordStart(pos) && memcmp(pos + 1, &MPORT[0], 5 * 2) == 0)
+      if (*(pos + 1) == 'm' && keywordStart(pos) && memcmp(pos + 1, &MPORT[0], 5 * 2) == 0)
```
guybedford: We could then change this to `memcmp(pos + 2, &PORT[0], 4 * 2) == 0`.

src/lexer.c
```diff
@@ -84,11 +84,11 @@ static inline __attribute__((always_inline)) bool consumeToken (char16_t ch) {
...
         tryParseImportStatement();
       break;
     case 'c':
-      if (keywordStart(pos) && memcmp(pos + 1, &LASS[0], 4 * 2) == 0 && isBrOrWs(*(pos + 5)))
+      if (*(pos + 1) == 'l' && keywordStart(pos) && memcmp(pos + 1, &LASS[0], 4 * 2) == 0 && isBrOrWs(*(pos + 5)))
```
guybedford: Same here.

guybedford (commented):
Looks good thanks, although seems wasteful to do the double compare if we can avoid it?

PR #224, opened 2026-07-07
perf: fast native UTF-16 source copy into Wasm memory on Node

guybedford:
Copying the JS source into Wasm linear memory is ~25% of total parse time, and it was done with a per-character `charCodeAt` loop (~1.2 GB/s). On Node we can instead encode the string as UTF-16 straight into memory natively via `Buffer` (~15 GB/s, ~13x), with no other lexer changes since the lexer already indexes UTF-16 code units.

Warm parse of the full sample corpus (3057 KiB, avg of 25 runs):

* before: ~11ms
* after: ~8ms  (~1.35x, ~27% faster end-to-end)

```ts
// before
copyLE(source, new Uint16Array(wasm.memory.buffer, addr, len));
// after
Buffer.from(wasm.memory.buffer, addr, len * 2).write(source, 'utf16le');
```

The whole gain comes from a single capability: encoding a JS string as UTF-16 directly into linear memory. Node exposes it natively; the browser has no equivalent (`TextEncoder` is UTF-8 only), so the portable `charCodeAt` path is retained as a fallback and the same speedup is left on the table anywhere a native UTF-16 encode-into-memory becomes available.

The fast path is gated strictly to the real Node `Buffer` on little-endian hosts (every platform Node runs on); browser bundles and big-endian hosts keep the existing code-unit copy. Existing tests pass unchanged, including the non-ASCII sample files.

### DIFF
--- a/src/lexer.ts
+++ b/src/lexer.ts
@@ -252,7 +252,12 @@ export function parse (source: string, name = '@'): readonly [
     wasm.memory.grow(Math.ceil(extraMem / 65536));
 
   const addr = wasm.sa(len - 1);
-  (isLE ? copyLE : copyBE)(source, new Uint16Array(wasm.memory.buffer, addr, len));
+  // Node's Buffer blits UTF-16 straight into Wasm memory ~10x faster than the
+  // charCodeAt fallback, in explicit LE matching Wasm regardless of host.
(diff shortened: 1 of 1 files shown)


BridgeAR (commented):
Nice!

PR #227, opened 2026-07-12
perf(lexer): skip identifier and number runs

BridgeAR:
## Summary

Skip the remainder of identifier and number runs after their first token dispatch, avoiding repeated switch dispatch and `lastTokenPos` stores.

On the 3,057 KiB sample corpus, Wasm improves from 15,356.31 to 13,010.46 us per sweep on Node 18.20.8 (-15.28%) and from 11,724.34 to 9,856.69 us on Node 24.18.0 (-15.93%). The asm.js build improves from 15,934.34 to 12,945.57 us on Node 24.18.0 (-18.76%). The full Wasm binary grows by 379 raw / 105 gzip bytes.

## Test plan

- [x] Run the full Wasm and asm.js test matrix, including minimal builds
- [x] Compare baseline and candidate metadata over 100,000 generated sources

### DIFF
--- a/src/lexer.c
+++ b/src/lexer.c
@@ -70,6 +70,12 @@ static inline __attribute__((always_inline)) bool handleSlash () {
   return false;
 }
 
+static inline __attribute__((always_inline)) bool isTokenRunChar (char16_t ch) {
+  // Fold ASCII case; NBSP is the only non-ASCII whitespace recognized here.
+  return (char16_t)((ch | 32) - 'a') < 26 || (char16_t)(ch - '0') < 10 ||
+    ch == '$' || ch == '_' || ch == '\\' || (ch > 127 && ch != 160);
+}
+
 // Consume one token at the current ch/pos, updating the global tokenizer state.
 // The single source of tokenization: the main loop and skipExpression both call
 // it, so the regex/keyword/import rules never diverge. Comments do not advance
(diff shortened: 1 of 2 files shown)


src/lexer.c
```diff
@@ -160,6 +170,12 @@ static inline __attribute__((always_inline)) bool consumeToken (char16_t ch) {
       openTokenStack[openTokenDepth++].token = Template;
       templateString();
       break;
+    default:
+      if (!isTokenRunChar(ch))
+        break;
```
guybedford, line 175:
```suggestion
```

Can't we just skip this, or is there a meaningful performance impact?

src/lexer.c
```diff
@@ -160,6 +170,12 @@ static inline __attribute__((always_inline)) bool consumeToken (char16_t ch) {
...
+    skipTokenRun:
+      while (isTokenRunChar(*(++pos)));
+      pos--;
```
guybedford:
```suggestion
      while (isTokenRunChar(*(pos + 1))) pos++;
```

guybedford (approved):

src/lexer.c
guybedford: Ahh, this is the current char, while the check is on the next char. nvm.

PR #228, opened 2026-07-20
Fix parsing of `export abstract class` syntax

mattwiller:
The current logic in `lexer.c` mis-parses code like `export abstract class Foo {}`, which we started observing in our builds when [import-in-the-middle@3.3.0](https://github.com/nodejs/import-in-the-middle/blob/main/CHANGELOG.md#330-2026-07-02) recently began parsing `.ts` files using this library.  This fix should resolve the issue by adding specific handling for the required syntax

### DIFF
--- a/lexer.js
+++ b/lexer.js
@@ -355,8 +355,22 @@ function tryParseExportStatement () {
       addExport(pos, pos + 7, -1, -1);
       return;
 
-    // export async? function*? name () {
     case 97/*a*/:
+      // export abstract class name ...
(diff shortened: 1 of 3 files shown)


BridgeAR:
Import in the middle is actually type stripping everything and if that does not work, it seems it is an issue in iitm, not here. es-module-lexer does not yet support TS, while it is planned for v3.

After looking into it closer, it seems like tsx is manipulating something on the fly to pretend to not be TS anymore (which kind of makes sense). That manipulation prevents iitm from understanding that it needs to type strip.

BridgeAR:
I am closing this, since it is fixed in iitm directly

PR #229, opened 2026-07-23
feat: add export origin analysis

BridgeAR:
> Rebased on `v3` with #204 and #205 landed, so the diff is now this feature alone. Supersedes #215 and #222. Adds the missing capacity bound on the template-span allocation so every arena write goes through the same grow (Wasm) / retry (asm.js) path.

## Summary

Detached exports can precede their imports, so source-order tracking cannot classify them or preserve their module request. This adds a full-build tagged union for direct exports, named and namespace reexports, and bare star reexports. Reexports link to the existing imports array through `fi`.

The minimal build keeps its v2 output and compiles the analysis path out.

## Why

Nothing about a binding is stored while parsing. A module that ends the parse with an unresolved export re-reads its import clauses once in the finalize pass, resolving names against a hash table of the pending exports; a module without one never builds the table and never reads a clause twice. That is the shape suggested in review, and it matches how real code is written: across 878 ESM files in two `node_modules` trees, 84.6% have imports and no clause export and pay nothing for the feature, while the 5.7% that do run the finalize walk carry a median of 2 import statements.

Three properties keep the common paths off the analysis:

1. The main loop scans an import clause for its closing brace rather than tokenizing specifiers, because only the finalize pass needs binding structure. This is why a module with imports and no clause export parses no slower than `main`.
2. A module with no import clause skips the table outright, and pending names are hashed at table insert rather than at the clause.
3. Table buckets are chained rather than open addressed, so one local exported under many names stays linear.

Analysis records live in the existing arena. The reservation is unchanged from `main`, and the arena grows only if a parse actually needs more.

Parsing those 878 real ESM files through `parse()` measures −1.1%, +0.6% and −1.1% against `main` over three runs — no measurable cost — and the 93 files that run the finalize walk land in the same band.

Synthetic shapes, median of five ABBA rounds against `main` on Node 24 / V8 13.6:

| input | vs main |
|---|---|
| real-world bundles (angular, d3, rollup, magic-string) | +0.1% geometric mean, nothing above the ±4 round spread |
| 400 imports plus direct exports | +1.2% |
| barrel of 400 `export … from` | +4.2% |
| 120 imports, 400 exports, trailing clause export | +8.2% |
| 400 detached reexports | +20.5% |

The last two rows are the trade this design takes: they carry far more imports per clause export than real modules do, so they pay the clause re-read at a rate real code does not reach. Small modules carry a fixed cost of roughly 20 ns per parse — 187 → 207 ns at the Wasm boundary for `import { a } from 'dep'; export const b = a; export { a as c };`.

Artifact sizes: `lib/lexer.wasm` 11,800 → 14,991 bytes and `dist/lexer.asm.js` 28,179 → 35,585. The minimal builds grow by 117 and 248 bytes.

Fixes: https://github.com/guybedford/es-module-lexer/issues/76
Refs: https://github.com/nodejs/import-in-the-middle/pull/259


guybedford:
I've pushed two maintainer commits to this branch:

* `ci:` CI was only wired to `main`, which is why no checks were running here — the workflow now also triggers for `v3` pushes and PRs, and the build is green on this branch (the full emsdk/fastcomp build, test, footprint and bench steps all ran).
* `fix:` the optional strip patterns in `build/combine-asm.mjs` now carry anchors — a pattern miss only passes silently when the function is genuinely absent from the emcc output after `EMSCRIPTEN_END_FUNCS`, while a still-present export with a stale regex throws. Also dropped `ess`/`et`/`eii`/`eit` from the `wasm` type in `src/lexer.ts` (they are only exported by the asm.js builds now) and noted `es`/`ee`/`els`/`ele` as minimal-build-only.

I also updated the PR description — the "binding capture is gated" paragraph described a previous revision, so it now reflects the unconditional inline collection.

guybedford:
On the allocation design: I don't think the `ImportBinding` records need to exist at all. The `Import` record already carries the statement range, so binding capture can be deferred to the end of the parse and gated on `pending_export_count != 0`: walk the import list, re-scan each static import clause bounded by the stored statement ranges, and resolve each binding against the pending-export table as it is read — streamed rather than stored.

That removes the `ImportBinding` struct, its list and its arena traffic, and takes binding collection off the hot path entirely, which should recover the 5.08 -> 5.28us regression on plain imports — modules without detached exports would pay nothing. Modules with them pay one bounded second read of just the clause text. I realize an earlier revision had a prescan that was dropped for cost — this is a different shape: no speculative forward scanning, only a deferred re-read of import clauses, and only when a detached export actually exists.

Beyond that, I want to keep the v2 allocation model as a hard constraint: static allocation with known limits and a known memory layout — no dynamic growth, no bail-and-retry. So `ensureAnalysisCapacity`, the wasm `memory_grow` path, `sal()` and the asm.js `-1` retry should all go. With `ImportBinding` gone the worst case becomes a knowable constant: a 32-byte `Export` plus an 8-byte bucket slot per 2-char clause name is ~20 bytes per source char, with imports well under that. The JS wrappers can reserve `source.length * K` for the analysis region up front (wasm grows once before the parse; asm.js folds it into `memBound`) and the whole growth machinery disappears. Note v2's implicit 2 bytes/char budget is already violable today by dense export clauses (~12 bytes/char), silently running into the asm.js stack region — sizing by the true constant fixes that too. K can be shaved if the constant feels heavy (u32 bucket indices instead of pointers, tighter `Export` packing).

The remaining per-record cost then looks right to me: +8 bytes on `Export` for classification, and the transient bucket table at 8 bytes per pending export (which could also fall back to a linear scan below a small threshold if we want to skip the table for typical modules).

It does require the clause reader to be callable outside the main loop (save/restore `pos`), but the collect functions are essentially already that.

guybedford (approved):

PR #233, opened 2026-08-09
feat: descriptive full-build record fields

guybedford:
This implements the descriptive full-build API for v3, following on from #229: the terse v2 field names and numeric type tags are replaced with properly typed discriminated unions.

* `Import = StaticImport | DynamicImport | ImportMetaRef`, discriminated by `type` with `phase: 'source' | 'defer' | null` replacing the overloaded numeric `ImportType` / `d` sentinel model
* descriptive fields throughout: `specifier`, `start` / `end`, `importStart` / `importEnd`, `dynamicStart`, `attributes` / `attributesStart`
* `Export = DirectExport | Reexport | ReexportAll` discriminated by `type: 'direct' | 'reexport' | 'reexport-all'` with `name`, `localName`, `importName`, `from`, `importIndex`, `exportStart`
* numeric enums leave the full-build API (`ExportType` is now internal to the wire format)
* the minimal build keeps the terse v2 shapes exactly, with its generated types now self-contained
* README split into Full Build | Minimal Build sections with all examples ported

Before:

```js
imports[0].n; imports[0].d === -2; exports[0].t === 2 && exports[0].f;
```

After:

```js
imports[0].specifier; imports[0].type === 'import-meta'; exports[0].type === 'reexport' && exports[0].from;
```

The shared suites assert through a strict terse mapping in the test harness, with the new test/full-api.cjs asserting the full record shapes directly and test/types.ts covering both unions at the type level. dist/lexer.minimal.asm.js is byte-identical before and after; the min wasm wrapper differs only by no longer exporting the v3-only `ExportType` enum.


BridgeAR (approved):
I had an in depth review and this is LGTM (both automated as well as manual)!

Benchmarks showed pretty much no regression using the descriptive fields, so that is also great (up to 1% slower in a worst case).

I also like the README updates that allow easy migration and how they reflect the current implementation better.

PR #235, opened 2026-08-10
perf: localize SIMD to profitable scan paths

BridgeAR:
SIMD setup dominates short scans, and whole-module `-msimd128` also changes scalar hot paths. This keeps short and unrelated scans scalar, then uses SIMD for long line comments and templates. Non-SIMD builds remain byte-identical to v3.

Long line-comment scans improve 9-65%, and long templates improve 66.8% against the SHA-identical scalar control on Node 18 arm64. The complete 3.06 MB sample corpus shows no aggregate regression across Node 18-26. Two isolated-process runs each produced a 0.94% equal-version geomean improvement. The SIMD build is 563 bytes smaller than #231.

### DIFF
--- a/chompfile.toml
+++ b/chompfile.toml
@@ -135,7 +135,7 @@ run = """
 	${{ EMSDK_PATH }}/emsdk install 6.0.0
 	${{ EMSDK_PATH }}/emsdk activate 6.0.0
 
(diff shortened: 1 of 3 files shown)


guybedford:
Thanks for working through these cases carefully!

Issue #155, opened 2023-07-31
`parse` returns a promise although the type does not indicate so


altinokdarici:
In [parse.ts](https://github.com/guybedford/es-module-lexer/blob/d44ad4ae1f5493a6956e226668008c9b2cd7f3fd/src/lexer.ts#L160C13-L160C13) file, the blow code snippet returns a promise. 

```ts
 if (!wasm)
    // actually returns a promise if init hasn't resolved (not type safe).
    // casting to avoid a breaking type change.
    return init.then(() => parse(source)) as unknown as ReturnType<typeof parse>;

```

This type hack causes issues on the consumption side. We call the parse function without awaiting as it's indicated by type system. However, we figured that parse may return Promise in the run time.

guybedford:
The function is supposed to be synchronous if you've awaited init, but if you haven't awaited init it should be async.

I think this was changed in the TypeScript migration. I personally like having the sync version, so would suggest we change the code to match the types like:

```js
if (!initialized)
  return init.then(() => ...);
// ... parse
```

guybedford:
Fixed in #237.

PR #236, opened 2026-08-16
fix: forward the sourcename through the pre-init parse path

hdimer:
When `parse()` is called before `init` resolves, it recurses as `parse(source)` and drops `name`, which then falls back to its `'@'` default. The README documents the second argument (`parse(source, 'optional-sourcename')`) and it documents calling parse before init ("either await init, or call parse asynchronously"), but the two don't compose:

```js
parse('import{', 'my-file.js')
// after awaiting init:  Parse error my-file.js:1:11
// before awaiting init: Parse error @:1:11
```

So anyone lexing a batch of files on the async path gets every error labelled `@`.

The fix is to pass `name` through the recursion. Only the wasm builds have this branch — `src/lexer.asm.js` and the plain JS build parse synchronously from the first call.

The test goes in the `Invalid syntax` suite behind the `WASM` gate. It imports a second module instance via a query string, since the suite's own instance has already awaited `init` and can't reach the branch. Verified it fails with `Parse error @:1:11` without the fix. `test:wasm` and `test:minimal:wasm` pass (150 each).

I left the `as unknown as` cast on that line alone deliberately — that's #155 and it needs a breaking type change, which didn't belong in the same diff.

Disclosure: I used an AI assistant while working on this. I reproduced the bug, reviewed the change, and ran the test suite myself.

### DIFF
--- a/src/lexer.ts
+++ b/src/lexer.ts
@@ -242,7 +242,7 @@ export function parse (source: string, name = '@'): readonly [
   if (!wasm)
     // actually returns a promise if init hasn't resolved (not type safe).
     // casting to avoid a breaking type change.
-    return init.then(() => parse(source)) as unknown as ReturnType<typeof parse>;
+    return init.then(() => parse(source, name)) as unknown as ReturnType<typeof parse>;
 
   const len = source.length + 1;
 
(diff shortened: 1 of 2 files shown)


BridgeAR (approved):
LGTM

PR #238, opened 2026-08-17
feat: always validate static specifier unicode in full build

guybedford:
This updates `StaticImport.specifier` so that it is now always a string; invalid escapes throw a positioned Parse error instead. All suites pass (wasm, asm, minimal × both, types).

What surfaced along the way: the asm (CSP) builds were already strict — src/lexer.asm.js has an eval-free acorn-style decoder that validates escapes and throws — only the eval-based wasm wrapper was lenient. So this change actually unifies a pre-existing build divergence rather than just tightening types:

- decode in src/lexer.ts now evals with a "use strict" prologue (matching the asm decoder in also rejecting legacy octal escapes like '\01', which sloppy eval silently accepted as \x01) and throws Parse error <name>:<line>:<col> with idx on failure — same message shape as the asm build. Applies uniformly: specifiers, string export names, and attribute keys/values, in full and minimal builds alike.
- DynamicImport.specifier stays string | undefined — undefined now exclusively means "not statically analyzable" (non-literal argument, non-glob template).
- Dropped the two as string casts on from: that the old type forced.
- README + doc comments updated; new test covers \uZZ, out-of-range \u{FFFFFF}, string export names, octal, and attributes (attributes gated !min since minimal never decodes them).

One residual asymmetry, deliberate: the asm decoder reports the exact offending character's position, while the eval path can only report the string's start (eval doesn't say where it failed). The test asserts the message shape without pinning the column.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 
 [![Build Status][actions-image]][actions-url]
 
-A JS module syntax lexer used in [es-module-shims](https://github.com/guybedford/es-module-shims).
+A JS/TS module syntax lexer used in [es-module-shims](https://github.com/guybedford/es-module-shims).
 
-Outputs the list of exports and locations of import specifiers, including dynamic import and import meta handling.
+Outputs the list and locations of exports and import specifiers, including dynamic import and import meta expressions.
 
-Supports new syntax features including import attributes and source phase imports.
+Supports new syntax features including import attributes, deferred evaluation, and source phase imports.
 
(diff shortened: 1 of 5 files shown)


BridgeAR (commented):
A few entries will still be accepted while they are also invalid such as `\u{}`, `\u0_00`. I guess while being on it, we could also fix these entries. That would be handled in readHexChar to throw a SyntaxError instead of just calling using break to stop the loop.

And we could strengthen the type assertion being correct by testing `imported.specifier satisfies string` in the type tests. I seems like a nice to have.

guybedford:
Thanks good points, added.

PR #248, opened 2026-09-07
perf: reduce parser wrapper overhead

BridgeAR:
Skip import-attribute reader work for records without attributes and reuse the Wasm source view below the existing Buffer threshold.

On Node 22.23.2, 20-pair CPU-time runs across 8-12 fresh processes reduce full asm.js corpus time by 5.2%. Wasm static-import workloads improve by 5.6-7.1%, and short-source workloads improve by 5.4-8.3%.

The cached view refreshes after memory growth and clears when asynchronous initialization replaces the synchronous instance.

### DIFF
--- a/src/lexer.ts
+++ b/src/lexer.ts
@@ -470,8 +470,12 @@ export function parse (source: string, name = '@'): readonly [
   // Buffer setup is slower than the loop for short sources.
   if (source.length >= 64 && hasBuffer)
     Buffer.from(wasm.memory.buffer, addr, (len - 1) * 2).write(source, 'utf16le');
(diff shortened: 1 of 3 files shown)


src/lexer.ts
guybedford, line 465: This is the only line that will trigger a memory grow right (we don't have malloc) - so just update the source view both here, and on init, then you can skip having to do the check on each post-parse.

guybedford: Ahh we do call memory grow in the full build now, nvm.

guybedford (approved):

